PR #3, opened 2014-09-19
Initial Pull Request

khandelwal:
This is the an initial pull request to set up peacecorps-site, and start a donation payment form. It's got a lot of things missing, but in the interest of having digestible pull requests I'm submitting this one. 

This still needs: 
- A better README
- More tests
- A country related field on the form
- Better form validation. 

But the rest of it should be ok, and is ready for review.


cmc333333:
I don't think you want to commit `local_settings.py`

peacecorps/peacecorps/templates/donations/donation_payment.jinja
```diff
@@ -0,0 +1,162 @@
...
+                {{form.card_dedication}}
+            </div>
+
+            <div>
+                {{form.dedication_consent.errors}}
+                <label for='id_dedication_consent'>Information Consent</label>
+
+                <ul id="id_dedication_consent">
+                <li><label for="id_dedication_consent_0">
+                    <input checked="checked" id="id_dedication_consent_0" name="dedication_consent" type="radio" value="yes-dedication-consent" />
```
cmc333333, line 116: I don't understand why these are hand-coded?

peacecorps/peacecorps/tests/test_forms.py
```diff
@@ -0,0 +1,19 @@
...
+            'street_address': '1 Main Street',
+            'city': 'Anytown',
+            'state': 'MD',
+            'zip_code': '20852',
+            'payment_type': 'credit-card',
+            'information_consent': 'vol-consent-yes'
+        }
+        form = DonationPaymentForm(data=form_data)
+        form_validity = form.is_valid()
+        self.assertEqual(form.is_valid(), True)
```
cmc333333: Obviously not important, but there's a `self.assertTrue(form.is_valid())`

peacecorps/peacecorps/templates/donations/donation_payment.jinja
khandelwal, line 116: I handed coded these in the assumption that when we work on the front-end we'll want to change the language. So, these got handcoded so that it was obvious that the language in the labels or how it's displayed could be easily changed.

khandelwal:
I think I've responded to everything.

PR #4, opened 2014-09-26
Changes to donation form, plus a simple donation review page

khandelwal:
This is a larger pull request than I would normally submit. Mostly, I forgot there were all these changes. 

This adds some missing fields to the donation form, and makes an attempt to break apart individuals and organizations. I'm sure in the future those two entities will see different fields too. 

Happy to answer questions, and look forward to the code review.


peacecorps/peacecorps/forms.py
```diff
@@ -1,19 +1,15 @@
 from django import forms
+
 from localflavor.us.forms import USStateField
 from localflavor.us.forms import USStateSelect
+from django_countries.widgets import CountrySelectWidget
```
cmc333333: Doesn't look like this gets used, and I don't think it's part of requirements.txt?

peacecorps/peacecorps/tests/test_forms.py
```diff
@@ -1,20 +1,39 @@
...
+            'street_address': '1 Main Street',
+            'city': 'Anytown',
+            'country': 'USA',
+            'state': 'MD',
+            'zip_code': '20852',
+            'payment_type': 'credit-card',
+            'information_consent': 'vol-consent-yes'
+        }
+        form = OrganizationDonationForm(data=form_data)
+        print(form.errors)
```
cmc333333: Stay debug statement

peacecorps/peacecorps/views.py
```diff
@@ -1,15 +1,90 @@
...
+    memo += phone
+    memo += '(yes)(no)(yes)'
+    return memo
+
+
+def donation_payment_review(request):
+    """ This view is for a simple donation payment review page. """
+    data = {}
+    for k, v in request.session.items():
+        data[k] = v
```
cmc333333, line 78: Ultimately, this needs to be saved into a model if it needs to be passed to pay.gov

cmc333333:
Any chance you can setup Travis? I think it would have caught the missing requirement.

peacecorps/peacecorps/views.py
khandelwal, line 78: True. I avoided that for now in the hopes that we can store less and less on our end.

khandelwal:
Ok, I think I've setup Travis. We'll see once this is merged right? 
Also, all other comments resolved.

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: python
+python:
+    - "3.4"
+    install:
+        - pip install -r requirements.txt --use-mirrors
+    script:
+        - python manage.py test --settings=peacecorps.settings.test
```
cmc333333, line 7: Whoops. Missed this before merging: lines 4 through 7 should all be outdented once, right?

khandelwal, line 7: Try: https://github.com/18F/peacecorps-site/pull/5

PR #5, opened 2014-09-26
Fix format of travis config file. 

khandelwal:


cmc333333:
Looks like you need a `cd`

khandelwal:
cd has been added.

PR #6, opened 2014-10-01
Neat + Sass instructions

cmc333333:
Sets up sass + bourbon + neat. Explicitly does not include these libraries in the repo, but adds instructions for how to set them up in the README.

If you're not interested in the front end, the compiled CSS is also checked in.

Nothing super crazy changes about the donations page -- just a little aligning to the grid.


seanherron:
+1

PR #7, opened 2014-10-03
Memo and custom field generation

cmc333333:
Match the fields that are generated in the current donations page.

### DIFF
--- a/peacecorps/peacecorps/forms.py
+++ b/peacecorps/peacecorps/forms.py
@@ -31,6 +31,7 @@ class DedicationForm(forms.Form):
     dedication_type = forms.ChoiceField(
         widget=forms.RadioSelect, choices=DEDICATION_TYPE_CHOICES,
         initial='in-honor', required=False)
+    #   @todo dedication_contact (present if "in memory")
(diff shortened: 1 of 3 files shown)


khandelwal:
Can you explain how and when we would use generate_custom_fields? If everything was hooked up, how is this used?

cmc333333:
Neither the memo nor the custom fields (nor the billing info, etc.) are processed by pay.gov when the browser is sent over. Pay.gov only accepts the agency id, tracking id, and app name from the browser.

It uses that to make a request to our server, at which point we provide the memo, custom fields, billing into, etc. So both the agency memo and the custom fields will need to operate on the stored model. In the meantime, just use the form's data.

cmc333333:
Alternatively, the model could have fields for memo/custom rather than storing the state of the form. We could go so far as to have the form generate the XML we need to respond with and have the model only store (tracking id, XML blob) pairs.

This would certainly make encryption easier.

khandelwal:
I think we need to talk more about this. However, that doesn't change the fact that this PR is good to go.

PR #8, opened 2014-10-04
Peace Corps Projects pages

annalee:
WIP. Not ready to merge yet.

Includes:
- [x] view and template for the donate landing page
- [x] models, view, and template for individual projects
- [x] models, view, and template for issues
- [x] sql dump of dummy data I added for testing

note: I did not add the test images I used, because I'm not sure of the copyright on some of them.

Still needs doing:
- [x] tests.
- [ ] make the templates presentable
- [ ] set media root, which determines where filefields upload to
- [ ] set image urls so they can be included in templates
- [ ] country fund pages
- [ ] general fund page
- [ ] special funds


peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+class CountryFund(models.Model):
+    country = models.ForeignKey('Country', related_name="fund")
+    fundcurrent = models.IntegerField(blank=True, null=True, default=0)
+    fundtotal = models.IntegerField(blank=True, null=True, default=10000)
+
+
+class FeaturedIssue(models.Model): 
+    issue = models.ForeignKey('Issue')
+
+    # Much like the Highlander, there can be only one.
```
cmc333333, line 27: :smile:

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+        return '%s (Featured)' % (self.project.title)
+
+
+class Fund(models.Model):
+    name = models.CharField(max_length=120)
+    description = description = models.TextField(blank=True, null=True)
+    featured_image = models.ForeignKey('Media',
+        help_text="A large landscape image for use in banners, headers, etc")
+    fundcurrent = models.IntegerField(blank=True, null=True, default=0)
+    fundtotal = models.IntegerField(blank=True, null=True, default=10000)
```
cmc333333: It would be useful to note that these are measured in cents.

I'm not sure I follow why `CountryFund`, `Fund` and `Issue` have these fields -- I thought we only showed hard numbers for `Projects`?

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+    issues_related = models.ManyToManyField(
+        'Issue', related_name="related_projects",
+        help_text="other issues this project relates to.")
+    featured_image = models.ForeignKey(
+        'Media',
+        help_text="A large landscape image for use in banners, headers, etc")
+    media = models.ManyToManyField(
+        'Media', related_name="projects", blank=True, null=True)
+    fundcurrent = models.IntegerField(blank=True, null=True, default=0)
+    fundtotal = models.IntegerField(blank=True, null=True, default=10000)
```
cmc333333: Should these ever be blank or null?

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+
+class Volunteer(models.Model):
+    HE = "H"
+    SHE = "S"
+    THEY = "T"
+    PRONOUN_CHOICES = (
+        (HE, 'He'),
+        (SHE, 'She'),
+        (THEY, 'They'),
+    )
```
cmc333333, line 159: I like this implementation

cmc333333:
Would it be better to use a fixture than a psql dump? It'd be more portable. It'd also allow us to separate the critical data (e.g. country names) from the dummy data.

peacecorps/peacecorps/views.py
```diff
@@ -88,3 +91,50 @@ def donation_payment_review(request):
             'app_name': settings.PAY_GOV_APP_NAME,
             'oci_servlet_url': settings.PAY_GOV_OCI_URL,
         })
+
+
+def donate_landing(request):
+
+    featuredprojects = [x.project for x in FeaturedProjectFrontPage.objects.all()]
```
cmc333333, line 156: Using `select_related()` should reduce the number of queries from n+1 to 1.

peacecorps/peacecorps/views.py
```diff
@@ -88,3 +91,50 @@ def donation_payment_review(request):
...
+
+    return render(
+        request,
+        'donations/donate_landing.jinja',
+        {
+            'featuredissue': FeaturedIssue.objects.get(id=1).issue,
```
cmc333333: Ideally, this could be written to account for no featured issues (e.g. the start state) without causing an explosion

peacecorps/peacecorps/views.py
```diff
@@ -88,3 +91,50 @@ def donation_payment_review(request):
...
+        {
+            'issue': issue,
+            'featured': featured,
+            'projects': projects,
+        })
+
+def donate_project(request, slug):
+
+    project = Project.objects.get(slug=slug)
+    # Passing this in separately to prevent multiple DB queries.
```
cmc333333, line 190: Django should cache the result after it's queried once. I believe using `select_related` will reduce the number of queries from two to one, though.

cmc333333:
This is a good start. I vote it be merged before everything else in the list is complete.

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+    code = models.CharField(max_length=5)
+    name = models.CharField(max_length=50)
+
+    def __str__(self):
+        return '%s (%s)' % (self.name, self.code)
+
+class CountryFund(models.Model):
+    country = models.ForeignKey('Country', related_name="fund")
+    fundcurrent = models.IntegerField(blank=True, null=True, default=0)
+    fundtotal = models.IntegerField(blank=True, null=True, default=10000)
```
khandelwal: What is this 10000 default?

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class Issue(models.Model):
+    name = models.CharField(max_length=100)
+    tagline = models.CharField(
+        max_length=140, help_text="a short phrase for banners (140 characters)")
+    call = models.CharField(
+        max_length=40, help_text="call to action for buttons (40 characters)")
+    description = models.TextField(blank=True, null=True)
+    slug = models.CharField(
```
khandelwal: Should this be a SlugField?

peacecorps/peacecorps/models.py
```diff
@@ -0,0 +1,164 @@
...
+        return '%s' % (self.name)
+
+
+class Issue(models.Model):
+    name = models.CharField(max_length=100)
+    tagline = models.CharField(
+        max_length=140, help_text="a short phrase for banners (140 characters)")
+    call = models.CharField(
+        max_length=40, help_text="call to action for buttons (40 characters)")
+    description = models.TextField(blank=True, null=True)
```
khandelwal: Shouldn't this be required?

khandelwal:
I think of the things to do: these things are important: 
set media root, which determines where filefields upload to
set image urls so they can be included in templates

And the rest we can do as a team later. I'm happy to help out.

annalee:
So the only reason I haven't done those two is that I've never had to deal with serving static files in production before and I'm confused about how to get it set up correctly (I'm probably overthinking it). If someone wants to pair with me on that, let me know.

khandelwal:
@cmc333333 @annalee What do you guys think of merging this and then following up on the rest of the stuff. I don't _think_ that would break anything. I'd rather move on to smaller pull requests.

annalee:
Should we wait until I've got some tests in here? I'm generally in favor of the idea of requiring at least basic test coverage before we merge, and I've got time to work on tests today.

khandelwal:
Ok that's fair. I'd say: 
1. We need some tests (maybe for the bits you consider the most complex). We can add more later. 
2. Now that I think about it, I'd like to see the SQL dump replaced with a fixture. That should be easy enough right?

annalee:
Yup. I'll work on those today.

cmc333333:
Great start, thanks @annalee

PR #9, opened 2014-10-06
Initial JS for donations form

cmc333333:
Adds JS form to form so that it matches the current donations page (including showing/hiding of fields and swapping of text).
Adds a missing field for dedication contact (which only appears if "in memory").
This does _not_ merge the individual + organization forms, though that'll be the next step.


peacecorps/peacecorps/templates/donations/donation_payment.jinja
```diff
@@ -14,22 +14,23 @@
...
             {{form.organization_name.errors}}
-            <label for="id_organization_name">Organization Name *</label>
+            <label for="id_organization_name" class="required">Organization Name</label>
             <div class="value">{{form.organization_name}}</div>
         </div>
 
         <div>
             {{form.organization_contact.errors}}
-            <label for="id_organization_contact">Organization Contact*</label>
+            <label for="id_organization_contact" class="required">Organization Contact</label>
```
khandelwal, line 27: The organization stuff isn't required if the donor type is individual. Is this going to be handled later when you combine things? Do you want me to combine things into one form?

cmc333333, line 27: Combining them into a single form is the next step on my list.

PR #10, opened 2014-10-07
Adding in the donation amount and project code

khandelwal:
This adds in the donation amount and project code as GET parameters to the page where donor information is entered. 

I use GET parameters so that the page can be bookmarked, passed around, etc. 

Some of the things in here will change, but I want to get this PR in so that we could move forward.


peacecorps/peacecorps/views.py
```diff
@@ -9,8 +9,25 @@
...
+    format. """
+
+    amount_dollars = amount_cents/100.0
+    return "$%.2f" % (amount_dollars)
+
+
 def donation_payment_individual(request):
     """ This is the view for the donations contact information form. """
+
+    amount = int(request.GET.get('amount', None))
```
cmc333333: This will 500 if the user entered a non-number

khandelwal:
@cmc333333 This has better error handling now. Merge worthy?

cmc333333:
Looks good, thanks!

PR #11, opened 2014-10-07
Combine forms. Add requirement validation checking

cmc333333:
- Combines the separate organization, individual, and dedication forms
- Adds validation checks to make sure conditionally required fields are met
- Adds JS to show/hide the organization/individual fields
- Runs all of the JS on page init. This is useful if the user has a validation error as the values on the form might not be their 'defaults' on page load in this case


peacecorps/peacecorps/forms.py
```diff
@@ -95,16 +94,38 @@ class DonationPaymentForm(forms.Form):
...
-    name = forms.CharField(label="Name *", max_length=100)
-
-
-class OrganizationDonationForm(DonationPaymentForm):
-    """ An actual organization form. This one is for organizations. """
-    organization_name = forms.CharField(
-        label='Organization Name *', max_length=40)
-    organization_contact = forms.CharField(
-        label='Contact Person *', max_length=100)
+    def required_when(self, guard_field, guard_value, check_field):
```
khandelwal, line 97: Ooh, I like this.

peacecorps/peacecorps/forms.py
```diff
@@ -6,45 +7,8 @@
...
-    dedication_email = forms.EmailField(label="Email", required=False)
-    dedication_address = forms.CharField(
-        label="Mailing Address", max_length=255, required=False)
-    card_dedication = forms.CharField(max_length=150, required=False)
-    dedication_consent = forms.ChoiceField(
-        widget=forms.RadioSelect, initial='yes-dedication-consent',
-        choices=DEDICATION_CONSENT_CHOICES, required=False)
-
-
 class DonationPaymentForm(forms.Form):
```
khandelwal, line 10: I guess this makes more sense. I was trying to separate out things that weren't necessary earlier.

peacecorps/peacecorps/urls.py
```diff
@@ -1,20 +1,16 @@
...
+from peacecorps.views import donation_payment
 from peacecorps.views import donation_payment_review
 
 urlpatterns = patterns(
     '',
     url(r'^admin/', include(admin.site.urls)),
     url(
-        r'^donations/contribute/individual$',
-        donation_payment_individual, name='donations_individual'),
+        r'^donations/contribute/?$',
```
khandelwal, line 11: What does the question mark do here?

cmc333333, line 11: It allows for the trailing slash to be present or absent. My understanding is that default django wants to insert those slashes, so we kind of wrestle with it when we don't allow the slash to be present (e.g. on eregs)

khandelwal, line 11: You've seen this:
https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-APPEND_SLASH
?

PR #12, opened 2014-10-09
Add migrations

khandelwal:
We just added a whole bunch of models. This adds in migrations, so that we start using migrations moving forward. 

selfie-0


cmc333333:
Whaaa!

PR #15, opened 2014-10-09
Remove sessions

khandelwal:
This removes the reliance on sessions to get to the review page. We're simply rendering a review page when the form validates correctly. 

This also makes a whole bunch of files PEP8 compliant.


cmc333333:
Thank you for removing the session

PR #14, opened 2014-10-09
PCPP template updates

cmc333333:
- Creates a base template
- Moved the donation form style into its own file
- Lots of style updates for PCPP - it looks closer to the balsamiq mockups now
- Bar chart added to project page

screen shot 2014-10-09 at 12 44 47 pm

Images aren't quite baked yet.

PR #16, opened 2014-10-09
Save/load images while developing

cmc333333:
This builds on #14.

While under development, use django to store and host images/media. Production will likely use `django-storages` or similar. Also adds some style tweaks so the images don't take over the page.

Media's stored and hosted from the "media" subdirectory of your working dir when you `runserver`.


peacecorps/peacecorps/models.py
```diff
@@ -115,8 +115,7 @@ def __str__(self):
         return '%s' % (self.title)
 
     def url(self):
```
khandelwal, line 118: Should we make this a property of the model?

cmc333333:
Made `url` a property - this makes the templates a little nicer, too.

@annalee let me know if this is holding you back on anything; I can separate out the file component from #14 and we can merge them separately.

khandelwal:
@cmc333333 We can easily get Theresa to audit this later. I'm happy to merge if it's holding anyone back.

annalee:
+1 on merge.

khandelwal:
Conflicts up in here.

cmc333333:
Merged upstream.

PR #17, opened 2014-10-11
Pay.gov XML Response Generation

khandelwal:
This generates the XML that we'll send Pay.gov upon request. This pull request simply generates the XML. It's not hooked up to URLs etc. 

The other major change in this pull request is naming of a few form fields for consistency. We know name form fields on the donor form by what they are called in the Pay.gov XML. 

Hopefully this doesn't cause any issues: @cmc333333


cmc333333:
Makes sense to me; does lxml need to be added to the requirements?

khandelwal:
I'm using Python's built in XML stuff.

PR #18, opened 2014-10-15
Fund model refactor

annalee:
Rather than storing funding info on project, issue, countryfund, etc models, we're moving it over to a separate Fund model and added a foreignkey from tables that are associated with funds. Updates to views, templates, migrations and test fixtures to use the fund model.


peacecorps/peacecorps/templates/donations/donate_project.jinja
```diff
@@ -25,15 +25,15 @@
   {% endif %}
   <div id="volunteer-info">
     <p><strong>{{project.volunteer.name}}</strong> from <strong>{{project.volunteer.homecity}}, {{project.volunteer.homestate}}</strong></p>
-    {% set funded_percent = project.fundcurrent * 100 / project.fundgoal %}
+    {% set funded_percent = project.fund.percent_funded() %}
     <div id="funded-chart">
       <div id="funded-amount-bar" style="max-width: {{funded_percent }}%;">&nbsp;</div>
-      <div id="funded-amount-text">{{ funded_percent|round|int }}% funded</div>
+      <div id="funded-amount-text">{{ funded_percent|int|round }}% funded</div>
```
cmc333333: If given 99.7, won't this give 99 rather than 100?

cmc333333: Maybe that's better? If so, let's just drop the `round`

PR #19, opened 2014-10-15
Add donation amount form on the project page

cmc333333:
There's a radio selector with four preset options (10, 25, 50, all) and a custom option. Not particularly styled.

### DIFF
--- a/peacecorps/peacecorps/forms.py
+++ b/peacecorps/peacecorps/forms.py
@@ -147,8 +147,8 @@ class DonationAmountForm(forms.Form):
                  ('preset-all', 'Fund the remaining amount')))
     # required if "custom" is selected above. Min value of $1, as anything
     # lower than that will cost too much money to process. Max value of
-    # $10,000, as anything above that can't be processed by pay.gov
-    payment_amount = forms.DecimalField(max_value=10000, min_value=1,
+    # $9,999.99, as anything above that can't be processed by pay.gov
+    payment_amount = forms.DecimalField(max_value=9999.99, min_value=1,
                                         decimal_places=2, required=False)
 
     def __init__(self, fund=None, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


peacecorps/peacecorps/forms.py
```diff
@@ -133,3 +134,41 @@ def clean(self):
...
+            if self.cleaned_data.get('presets') == 'preset-' + str(amt):
+                return Decimal(amt)
+        if self.cleaned_data.get('presets') == 'preset-all':
+            if not self.fund:
+                raise ValidationError('Missing fund')
+            else:
+                remaining_amount = self.fund.fundgoal - self.fund.fundcurrent
+                if remaining_amount < 100:  # cents
+                    raise ValidationError('Must be >= 1.00')
+                elif remaining_amount > 1000000:
```
khandelwal: Let's add a comment to explain why we're doing this.

peacecorps/peacecorps/tests/test_views.py
```diff
@@ -175,3 +174,30 @@ def test_issue_rendering(self):
     def test_project_rendering(self):
         response = self.client.get('/donate/project/test-project')
         self.assertEqual(response.status_code, 200)
+
```
khandelwal, line 177: Thank you for tests!

peacecorps/peacecorps/forms.py
```diff
@@ -133,3 +134,43 @@ def clean(self):
...
+        if self.cleaned_data.get('presets') == 'preset-all':
+            if not self.fund:
+                raise ValidationError('Missing fund')
+            else:
+                remaining_amount = self.fund.fundgoal - self.fund.fundcurrent
+                # We're circumventing the normal bounds checks, so we must
+                # enforce them here
+                if remaining_amount < 100:  # cents
+                    raise ValidationError('Must be >= 1.00')
+                elif remaining_amount > 1000000:
```
khandelwal: @seanherron says this should be >= 1000000

seanherron:
Question - can we show the actual amount you would be paying next to the "fund the remaining amount" option?

khandelwal:
So, in this we're making putting >= $10,000 as a validation error for the form. That's fine.  However, should we be handling that case more gracefully? Are there specific instructions we should direct the user towards for large donations?

Asking...for a friend.

cmc333333:
@seanherron @khandelwal - both great questions; bring them up with design.

I don't believe either concern should prevent this pull request, a first draft, from being merged.

PR #20, opened 2014-10-15
Add production settings file for wsgi

cmc333333:


khandelwal:
This is good to go - but let's see if @annalee has the right permissions.

PR #23, opened 2014-10-20
Save donor info

cmc333333:
First draft of saving donor info.
1. Creates a new application to store pay.gov-related functions - this division should allow us to limit which urls are accessible simply by turning applications on and off
2. Move payxml and several related functions from the view code into the new app
3. Save the generated xml when a user hits the confirmation page
4. Remove some unneeded fields from the review submission form


peacecorps/paygov/models.py
```diff
@@ -0,0 +1,17 @@
+from datetime import timedelta
+
+from django.db import models
+from django.utils import timezone
+
+
+def default_expire_time():
```
khandelwal, line 8: My suspicion is that we'll need more time than this for the default expiry. Should this become a setting so that we can change it easily?

peacecorps/peacecorps/settings/base.py
```diff
@@ -25,6 +25,7 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'peacecorps',
+    'paygov',
```
khandelwal, line 28: Hmm. Do we want to remove this out of the base setting? Putting it here means it gets included on all servers. Maybe there's a copy of the production settings for paygov servers?

khandelwal:
Some minor thoughts I'd like us to talk about. Otherwise this is good to merge!

peacecorps/paygov/models.py
cmc333333, line 8: Good idea - on it.

peacecorps/peacecorps/settings/base.py
cmc333333, line 28: You bring up a good point. It's made me realize that the paygov models can't live there (lest they not be created w/ migrations). Moving the models around (and keeping only the views) is a big enough change that I want to put it in a later pull request.

khandelwal, line 28: The migrations will run if you add in the application. I think it even warns you that there are migrations that haven't run. So I think we might be safe that way?

cmc333333, line 28: Right, but you won't be able to use the app _without_ paygov -- you won't be able to save donor info because the db won't be set up.

This won't be a problem in production, as individual instances share a db, but I'd prefer not to require that environment.

cmc333333:
Updated such that the expire time is now a setting

PR #24, opened 2014-10-20
First pay.gov callback

cmc333333:
This builds on #23, retrieving the model built there when pay.gov requests it. 

Logs all input as a debug message.

The interesting bit is this commit: https://github.com/cmc333333/peacecorps-site/commit/a069e882851de80adf63a865d0a99f7d96cef7d2


khandelwal:
Again - minor comments.

cmc333333:
Updated with content type and a 404

PR #25, opened 2014-10-21
adding tinymce

annalee:
TinyMCE requires static_root to be set. Also added some basic configs.


cmc333333:
@annalee - please add STATIC_ROOT to the test settings, too

annalee:
Added static_root to test settings. Tests are now passing.

PR #27, opened 2014-10-21
Second Pay.gov callback

cmc333333:
- Adds a new model to track individual donations
- Responds to pay.gov with error information during any of a half-dozen error conditions
- Otherwise, saves the donation, deletes the donor info, and responds with success

### DIFF
--- a/peacecorps/paygov/views.py
+++ b/peacecorps/paygov/views.py
@@ -1,11 +1,12 @@
 import logging
+import re
 
 from django.http import HttpResponse, HttpResponseBadRequest
 from django.http import HttpResponseNotAllowed
 from django.shortcuts import get_object_or_404
 from django.views.decorators.csrf import csrf_exempt
 
-from peacecorps.models import DonorInfo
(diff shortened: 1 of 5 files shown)


peacecorps/paygov/views.py
```diff
@@ -20,3 +21,36 @@ def data(request):
...
+        if not info:
+            message = 'Invalid agency_tracking_id'
+        else:
+            donation = Donation(
+                amount=float(request.POST.get('payment_amount'))*100)
+            donation.fund_id = info.fund_id
+            donation.save()
+            info.delete()
+            message = 'OK'
+    return HttpResponse('response_message=' + message,
```
khandelwal, line 55: You're sending 200 codes back even in error situations. Is that how the communication with pay.gov works? They don't accept any other codes?

cmc333333, line 55: It's unclear from their documentation. 200s are the safest bet until we test otherwise.

khandelwal:
Looks good to me.

PR #28, opened 2014-10-22
Simple success/failure pages

cmc333333:
The user is returned to these pages after their transaction. Just adding place holders for now.

### DIFF
--- a/peacecorps/peacecorps/templates/donations/failure.jinja
+++ b/peacecorps/peacecorps/templates/donations/failure.jinja
@@ -1,0 +1,4 @@
+{% extends "donations/base.jinja" %}
+
+{% block title %}Success!{% endblock %}
+{% block header %}Success!{% endblock %}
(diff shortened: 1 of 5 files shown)


peacecorps/peacecorps/templates/donations/failure.jinja
```diff
@@ -0,0 +1,4 @@
+{% extends "donations/base.jinja" %}
+
+{% block title %}Success!{% endblock %}
```
khandelwal: This should probably say Failure! (So that it doesn't get confusing when this page is loaded)

khandelwal:
This one requires a simple change on your part.

peacecorps/peacecorps/templates/donations/failure.jinja
cmc333333: :o

cmc333333:
Updated the text on this one; my bad.

PR #29, opened 2014-10-22
Removing the volunteer model

annalee:


cmc333333:
Note to @khandelwal and @theresaanna - this requires you clear out existing data and then reload

PR #34, opened 2014-10-23
Ensure password complexity requirements are met

seanherron:
As outlined in the System Security Plan, we need to set some more restrictive password requirements on users. This PR adds the `clean_password1` function to `UserCreationForm` and `AdminPasswordChangeForm` to check passwords against the `password_validator` before accepting them when using the Django admin panel.

There are additional requirements in terms of password reuse and renewal policies that still need to be implemented.


peacecorps/peacecorps/admin.py
```diff
@@ -1,7 +1,67 @@
...
+from django.contrib.auth.forms import (UserCreationForm,
+    AdminPasswordChangeForm)
+from django.contrib.auth.models import User
+from django.forms import ValidationError
+
 
 from peacecorps import models
 
+# Password length requirements
+def password_validator(password):
```
khandelwal, line 14: Why not have this just raise the ValidationError? That might clean up the rest of the code some.

peacecorps/peacecorps/admin.py
```diff
@@ -1,7 +1,67 @@
...
+        errors = password_validator(password)
+
+        # If password_validator returns errors, raise an error, else proceed.
+        if errors:
+            raise ValidationError('\n'.join(errors))
+        else:
+            return password
+
+class StrictAdminPasswordChangeForm(AdminPasswordChangeForm):
+    def clean_password1(self):
```
khandelwal, line 61: Should the help_text for the password field indicate all the rules? Or would you want to put those elsewhere?

peacecorps/peacecorps/admin.py
khandelwal, line 14: To might also want to put a note in the docstring to indicate why we're requiring these requirements.

seanherron, line 14: Indicated this in the DocString. I was getting some strange errors when trying to raise the ValidationError in the function as opposed to in the Form class so left it there, but can try again if it would make things clearer.

peacecorps/peacecorps/admin.py
seanherron, line 61: Good point, help_text added.

cmc333333:
Very nice. Thanks for being a good sport, @seanherron !

PR #39, opened 2014-10-24
GPG-encrypted XML field

cmc333333:
This creates a GPG field type and sets the xml field on donations to it. The field is backed by a binary blob in the DB, converted back into a string when needed.

GPG requires a home directory configured; if one is not, GPG will not be enabled, and all operations will operate on plain text. When migrating, you will receive a warning about this.

Individual fields can be configured with a specific recipient's key in mind, allowing different fields to be decrypted by different keys. In practice, the server with access to pay.gov will have a private key, whose corresponding public key is available to all of the other web servers.


seanherron:
+1

@cmc333333 you'll have to walk me through how we can set this up in deployment. What should the filenames of the keys be? I'm assuming we can keep them in a secure location (eg. S3 bucket) and deploy to instances during configuration.

PR #40, opened 2014-10-27
Add a second donation form to the project page

cmc333333:
The designs have a donation form at the middle and bottom of the page. Implemented this with form prefixes (lest the values overwrite each other).

Also removes the duplicate donation button.


khandelwal:
This looks good to me.

PR #41, opened 2014-10-27
New memorial and general fund pages.

annalee:
Also includes a bugfix for the gpg tests.


cmc333333:
This makes sense. Something to consider for a future iteration: many of these models now share common fields. It might make sense to abstract that info into a separate class to make our types match up.

annalee:
I'm actually thinking we can cut some of these models. For starters, MemorialFund can be folded into FundDisplay (which needs a more useful name).

PR #43, opened 2014-10-27
Created layout of main navigation

msecret:
Setup the layout part of main navigation, have not dealt with type
treatment yet.

Added sass component and globals folder, components to store all sass
modules and globals for global variables.

Added a util file for general utilities like clears, floats, etc. Added
an umarked list style there.

TODO: probably don't need a separate folder for globals, just one file.

Other changes:
- Setup grid settings as per Neat setup instructions:
  github.com/thoughtbot/neat
- Fixed some whitespace issues.

screen shot 2014-10-27 at 2 04 23 pm


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -5,7 +5,28 @@
...
+                  <a href="/about">About</a>
+                </li>
+                <li class="main_nav__link">
+                  <a href="http://passport.peacecorps.gov/">Blog</a>
+                </li>
+                <li class="main_nav__link">
+                  <a href="/volunteer">Volunteer</a>
+                </li>
+                <li class="main_nav__link">
+                  <a href="/donate">Donate</a>
```
cmc333333: Since we've got this page already build, you can use `{{ url("donate landing") }}` rather than hard-coding.

cmc333333:
So glad you're here @msecret

PR #45, opened 2014-10-28
Updating icon field on issue model

annalee:
Changing the icon field on the issue model to an FK to media, to make icons easier to re-use without duplicating.


peacecorps/peacecorps/migrations/0019_auto_20141028_1736.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('peacecorps', '0018_issue_icon'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='issue',
+            name='icon',
+            field=models.ForeignKey(help_text='A small icon to represent the issue on the landing page.', to='peacecorps.Media', related_name='iconissues'),
```
msecret, line 17: Do we have any guidelines on keeping lines to less the 80 chars long?

msecret:
LG to me, although maybe one of the backend developers should be the one that merges it.

peacecorps/peacecorps/migrations/0019_auto_20141028_1736.py
cmc333333, line 17: Generally, PEP8 recommends 80 chars, but this is file is auto-generated

cmc333333:
Something to keep in mind - fking to the generic "Media" model means that it's on the CSS/Markup to size appropriately.

cmc333333:
@msecret - note that this requires you wipe your db to upgrade

PR #46, opened 2014-10-28
Enhc full bg featured img

msecret:
This is a "featured" project image that spans from top of the page
til the next section.

Uses a new background image class and user applies a background url
along with a height inline.

Not 100% confident in my choice to use z-indexing on this solution, but
am also not 100% sure on how this bg image and other will be used
across the site.

Other changes:
- Setup z-index variables to avoid collisions.

screen shot 2014-10-27 at 4 44 23 pm


peacecorps/peacecorps/templates/donations/donate_landing.jinja
```diff
@@ -11,6 +11,11 @@
 {% block content %}
 
 {% if featuredissue %}
+<div class="full_bg" style="
+    background-image: url('http://lorempixel.com/1920/1200/people');
+    height: 46em;
+  ">
+</div>
```
cmc333333, line 18: Should this be the style for the `section` tag below (and remove the `img`)?

msecret, line 18: Yea this is true. I'm actually dealing with this in a separate branch that I've been working on that is dealing with the tagline and button. Just this be merged separately before I push that branch, or should I push that branch first?

cmc333333, line 18: Sure, we can merge and fix in a later PR.

PR #48, opened 2014-10-28
Script to synchronize payments with a CSV

cmc333333:
Adds a management script which takes a csv file name argument. Based on the values in the associated csv, transactions are deleted and fund amounts are updated.

Adds timezone configuration info (though we encourage UTC for everything).

Uses `round` when multiplying to turn dollars into cents to avoid annoying floating point issues.


peacecorps/peacecorps/tests/test_sync.py
```diff
@@ -0,0 +1,61 @@
+from datetime import datetime
```
msecret: I would have thought that this file would be named "test_sync_payments.py" to line up with the file it's testing, although I'm less familiar with our naming conventions

peacecorps/peacecorps/management/commands/sync_payments.py
```diff
@@ -0,0 +1,46 @@
...
+              Generally, this means deleting transactions and updating the
+              amount field in the fund."""
+
+    def handle(self, *args, **kwargs):
+        if len(args) == 0:
+            raise CommandError("Missing path to csv")
+
+        with open(args[0]) as csvfile:
+            # Column names will no doubt change
+            for row in csv.DictReader(csvfile):
```
msecret, line 38: Where are these CSV files coming from?

In my previous experience with python's csv lib, I ran into some trouble with certain csv dialects detecting new lines and different fields correctly. Are all the csvs uploaded going to be from the same program?

peacecorps/peacecorps/tests/test_sync.py
cmc333333: Can do, it would be more consistent.

peacecorps/peacecorps/management/commands/sync_payments.py
cmc333333, line 38: CSV files are being provided to us by the peace corps. We don't have an example yet, so this is testing blind.

cmc333333, line 38: I should rephrase - we have an older xls file (which is where field names and column formats came from), but not a CSV to address your concern.

PR #49, opened 2014-10-29
ench The featured project info and callout

msecret:
This is the large, orange callout for the featured project. It uses
a full width banner component along with the showcase component to
display itself as an orange banner with the content and callout button
centered.

Have not worked on button yet or spacing below this element.

Some issues I see at this point for discussion:
- The featured project image is a variable that the user (peace corps employee) can control. If a good image (good size and good color) is not chosen, this could break the layout, as you may not be able to see the navigation or the images will pixelated.
- We're not displaying the featured project's "description" anymore, which was part of the site before this work.

screen shot 2014-10-28 at 5 01 57 pm


ericronne:
My assumption was that we would have a hand in the creation of future hero content such as this. If instead we're putting this content generation fully into the hands of a peace corps employee, yes, we may need to revisit the look. ... Perhaps we could create a simple style guide of sorts?

PR #53, opened 2014-10-30
ehnc Seciton headings

msecret:
The large, half-cirle section starts with a title in the middle.

This implementation can work with text or other content in the middle.


cmc333333:
@msecret: I think you need to pull from upstream?

cmc333333:
Splendid - I like how you're breaking this up

PR #54, opened 2014-10-30
Campaign model refactor

annalee:
@cmc333333 and I went through and refactored our models to reduce the number of models that refer to funds. We've simplified them into two models: projects (for pcpps), and campaigns (for everything else). Campaigns can be groups of projects and/or they can be entities that collect money to distribute to projects.

For instance a memorial fund is not a group of projects, but collects money.
A country fund is a group of projects AND it collects money.
A campaign can also be used to 'tag' related projects WITHOUT collecting money.


cmc333333:
May the nightmares end here.

@msecret this is another one which requires a db wipe and reload, but it'll hopefully be the last.

PR #56, opened 2014-10-31
Project funded

cmc333333:
Button text changes and pay-the-remaining-amount option disappear when the project is funded.

If the form is submitted in this scenario (and the project has an overflow fund), the donation is directed to that overflow fund.

### DIFF
--- a/peacecorps/peacecorps/models.py
+++ b/peacecorps/peacecorps/models.py
@@ -136,7 +136,7 @@ class Fund(models.Model):
         return percentfunded(self.fundcurrent, self.fundgoal)
 
     def funded(self):
-        if self.fundcurrent >= self.fundgoal:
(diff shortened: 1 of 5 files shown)


peacecorps/peacecorps/models.py
```diff
@@ -136,7 +136,7 @@ def percent_funded(self):
         return percentfunded(self.fundcurrent, self.fundgoal)
 
     def funded(self):
-        if self.fundcurrent >= self.fundgoal:
+        if self.fundgoal is not None and self.fundcurrent >= self.fundgoal:
```
khandelwal: Should this just be:

`return self.fundgoal is not None and self.fund_current >= self.fundgoal`

cmc333333:
Thanks for the suggestion, @khandelwal. Updated

PR #58, opened 2014-10-31
Renaming "Fund" model to "Account"

annalee:
Changes the "Fund" model to "Account" to make it clear that the model only references the money, and not all the other things that go with a particular fund (which we store in Campaign or Project, depending on the fund type).

Sorry, I know this changes basically every file. But @CM and I think we have the models in stable shape after this change, so we shouldn't (famous last words) need to do any more of this kind of refactoring.


cmc333333:
I think this good to merge once the tests pass. @annalee - do you know if this will require dropping and reloading the db?

annalee:
I was able to delete the fund model and run the migrations without dropping and reloading the db.

cmc333333:
Let's get a data migration in so the server can just `migrate`

cmc333333:
Thanks @annalee, migrations work for me

PR #60, opened 2014-11-03
Allow allowed_hosts to come from environment

cmc333333:


peacecorps/peacecorps/settings/production.py
```diff
@@ -7,6 +7,9 @@
 
 SECRET_KEY = os.environ.get('DJANGO_SECRET_KEY', '')
 ALLOWED_HOSTS = ['localhost', '127.0.0.1']  # proxied
+if os.environ.get('DJANGO_HOST'):
+    ALLOWED_HOSTS.append(os.environ.get('DJANGO_HOST'))
```
khandelwal, line 11: Why wouldn't you call it DJANGO_ALLOWED_HOSTS? Or something that maps more clearly name-wise to ALLOWED_HOSTS?

seanherron, line 11: @khandelwal Because it is expecting a single value rather than multiple values I assume.

khandelwal, line 11: Then shouldn't it be: DJANGO_ALLOWED_HOST? I know it sounds pedantic, but our environment variables that correspond to django settings should match that name so that the mapping is obvious. Maybe it's obvious to y'all.

PR #61, opened 2014-11-03
Import PCPPs

cmc333333:
Extends the donation amount synchronization script to also account for project imports.
- Adds a `published` field on PCPPs, defaulting to False; data migration to move all existing entries to be published
- Allow featured_image to be blank on projects
- Adds a `published_objects` manager on PCPPs, which is used throughout the site
- When importing project information, look up the associated sectors/country/etc.
- If it can't find the sector/country referenced, it'll log a warning, but continue

Note that we may need to tweak the sector mapping once we have a better dataset.


msecret:
Concerning projects not requiring a featured image, if theres anything I should know about how to account for that on the front-end, let me know.

PR #64, opened 2014-11-04
Success/Failure pages specific to projects/campaigns

cmc333333:
Adds some sparse templates for success and failure pages after a donation. These must next be specified in the pay.gov interaction, overriding the default /success and /failure pages.


msecret:
I like how you separated out the templates!

PR #68, opened 2014-11-05
Adding pay.gov payments to account totals for display

annalee:


peacecorps/peacecorps/models.py
```diff
@@ -49,11 +49,14 @@ class Account(models.Model):
     def __str__(self):
         return '%s' % (self.name)
 
+    def total(self):
+        return self.current + sum([x.amount for x in self.donations.all()])
```
cmc333333: `sum(x.amount for x in self.donations.all())` is a little cleaner -- you're creating a generator rather than a list. 

Both options require loading all of the donation objects from the database. For our use case, that shouldn't be terrible, but it'd be better to leave that calculation in to database using aggregations.

cmc333333:
Great, thanks @annalee

PR #69, opened 2014-11-05
Modified production settings to allow from any hostname

seanherron:
This PR attempts modifying `ALLOWED_HOSTS` to allow from any domain name. Since in production we will be operating inside a private subnet with no access that we don't allow through the LB, this should still be a safe default.


cmc333333:
Sad but true.

PR #71, opened 2014-11-06
Start on project page

msecret:
Started on layout for project page. Also started on the difficult
percent funded circle, but didn't get to finish that so didn't put it
up.

Still have to put some volunteer pictures in my local db to test that
part of the layout.


peacecorps/peacecorps/templates/donations/includes/project-top.jinja
```diff
@@ -1,26 +1,42 @@
...
+  <section class="showcase--col3_centered u-clearfix">
+    <section class="showcase--col3_centered__col">
+      <p class="t-bold">{{project.title}}</p>
+      {% if project.volunteerpicture %}
+        <div class="">
+          <img src="{{project.volunteerpicture.url}}"
+               alt="{{project.volunteerpicture.description}}">
+        </div>
+      {% endif %}
+      <span>{{project.volunteername}} from {{project.volunteerhomestate}}</span>
```
cmc333333, line 25: Do we not need homecity? We don't get that data from the project imports, so it'd be great to drop it from the model if it's not in the comps.

PR #74, opened 2014-11-07
Fixtures for country, memorial, issues, and other campaigns

cmc333333:
Currently includes some non-final images and puts a few of the non-issue sector funds as "general", which may need to change.

Accounting codes are all up-to-date.

These fixtures can be used when populating the database for the first time. If you've already imported the tests.yaml fixture, you'll run into some conflicts.


msecret:
Is there specific instructions on how to install these? Do they require a db erase?

PR #79, opened 2014-11-07
Don't allow a user to reuse a password

cmc333333:
With test and PEP8 cleaning.

### DIFF
--- a/peacecorps/peacecorps/tests/test_admin.py
+++ b/peacecorps/peacecorps/tests/test_admin.py
@@ -2,8 +2,8 @@ from django.contrib.auth.models import User
 
 from django.test import TestCase
 
-from peacecorps.admin import (StrictUserCreationForm,
-StrictAdminPasswordChangeForm)
+from peacecorps.admin import (
+    StrictUserCreationForm, StrictAdminPasswordChangeForm)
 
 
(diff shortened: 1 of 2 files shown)


peacecorps/peacecorps/tests/test_admin.py
```diff
@@ -176,6 +181,16 @@ def test_passwordsuccess(self):
...
-        self.assertTrue(form.is_valid())
\ No newline at end of file
+        form = StrictAdminPasswordChangeForm(
+            data=form_data, user=User.objects.get(username='testuser'))
+        self.assertTrue(form.is_valid())
+
+    def test_new_password(self):
+        """ Password must be different than the current """
+        form_data = {
+            'password1': 'q*A9x=^2hg&v7u?u9tg?u',
```
msecret: On this test, where is this password that you're testing against originally set on the user account?

I would expect to see some setup code that sets this password on the account and then attempts to submit it.

cmc333333: That's in the `setUp` method. I'll rewrite to make that more clear, though.

cmc333333:
Updated to call out the fact that the test is reusing a password.

PR #80, opened 2014-11-08
Enhc svg map

msecret:
SVG map solution.

Uses an SVG map and JS to zoom in on the selected country and
highlight its fill color. Solution can scale to any country in the
known world.

Known problems:
- The gzipped, minified svg map size is still a little too large at
  367kb. We could investigate a smaller map or tweak this solution to use
  svg maps that are separated by regions or countries.

TODO:
- Add build process to minify svg files.
- Separate JS into separate files.

screen shot 2014-11-07 at 2 50 29 pm


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -42,8 +42,16 @@
...
 </div>
-{% block custom_js %}{% endblock %}
+<script type="text/javascript">
+  {# TODO JS should all be in its own file, for better organization, testing and
+     debugging. #}
+  window.onload=function() {
+    {% block custom_js %}{% endblock %}
+  };
+</script>
+
```
cmc333333, line 48: This would break the donation and landing page JS

peacecorps/peacecorps/templates/donations/includes/project-top.jinja
```diff
@@ -49,3 +56,53 @@
     </div>
   </div>
 </article>
+
+{% macro map_js(country_code) %}
```
cmc333333: Putting JS in the template like this is really messy. Let's put it in it's own .js file as pass the country parameter in via a data attribute or a single function call.

peacecorps/peacecorps/static/peacecorps/sass/_globals.scss
```diff
@@ -8,7 +8,7 @@
  * Color
  */
 $black: #222;
-$white: #fff;
+$white: #ffffff;
```
cmc333333, line 11: Curious, what's the benefit here?

peacecorps/peacecorps/views.py
```diff
@@ -144,6 +145,7 @@ def donate_project(request, slug):
         request,
         'donations/donate_project.jinja',
         {
+            'COUNTRY_CODE_MAP': COUNTRY_CODE_MAP,
```
cmc333333: No need to pass in the whole map, you can just pass in the two-letter code itself: `COUNTRY_CODE_MAP[project.country.code]`

peacecorps/peacecorps/countries.py
```diff
@@ -192,3 +192,253 @@
     ("YEM", "Yemen"),
     ("ZAM", "Zambia"),
     ("ZIM", "Zimbabwe"),)
+
+COUNTRY_CODE_MAP = {'SAU': 'SA',
```
cmc333333: This mapping is pretty onerous. @annalee or @khandelwal Do you remember if there's a specific reason for a three-letter code? The only thing I can think of is the potential to map to fund codes?

We could either switch to use two-character identifiers, or modify the SVG file to use the three-character codes.

peacecorps/peacecorps/static/peacecorps/sass/_globals.scss
msecret, line 11: I don't think there is one. I think this change occurred because I was copy and pasting things.

peacecorps/peacecorps/static/peacecorps/js/.jshintrc
```diff
@@ -0,0 +1,11 @@
+{
```
cmc333333, line 1: This seems like a better place. Would you mind merging in the contents of `peacecorps/.jshintrc` with this and deleting that file?

cmc333333:
Very cool

PR #84, opened 2014-11-12
Password expiration

cmc333333:
- Moves password checking code into a separate app devoted to the content editor interface
- Effectively extends the user model via a one-to-one relationship, which adds a password expiration time
- When a user changes password (through the existing form), that password expiration time is reset
- All admin pages (excepting a whitelist) are wrapped in a check that the user's password hasn't expired. If it has, they are redirected to the password setting form

Note that while working on this ticket, I discovered that the password checks built into the admin form can be circumvented by hitting a different url. I'll open a separate task for that.


peacecorps/contenteditor/urls.py
```diff
@@ -0,0 +1,44 @@
...
+                or (request.path ==
+                    '/admin/auth/user/%d/password/' % request.user.id)
+                or (request.user.extra.password_expires > timezone.now())):
+            return callback(request, *args, **kwargs)
+        else:   # Password expired and not in whitelist
+            # @todo: it makes more sense to not hard-code this, but we need to
+            # change password validation first
+            return HttpResponseRedirect(
+                '/admin/auth/user/%d/password/' % request.user.id)
+            # return HttpResponseRedirect(reverse('admin:password_change'))
```
msecret, line 23: This is commented out, is it needed?

PR #87, opened 2014-11-12
enhc The donate form with button UI

msecret:
Uses a new version of the button class to make square buttons to create
this interface. Still works as a form of radio buttons.

TODO
- Clicking on the custom amount input box will not trigger the radio
  option, may need JS to help with this.

@ericronne as you can see, I have not been messing with the fonts at this point because they are not finalized.

When selecting first payment amount:
screen shot 2014-11-12 at 2 46 12 pm

When entering custom amount:
screen shot 2014-11-12 at 2 46 21 pm

When at small breakpoint:
screen shot 2014-11-12 at 2 46 39 pm


peacecorps/peacecorps/static/peacecorps/css/donation.css
```diff
@@ -687,14 +698,58 @@ body {
...
+    .bubble__content, .bubble_form input[type=number]:not(old) {
       padding-top: 20%; } }
 
 .bubble--neutral_dark {
   background-color: #E5E1D4; }
 
+.bubble--primary, .bubble_form input[type=radio]:not(old):checked + .bubble {
+  background-color: #A12122;
+  color: #ffffff; }
+  .bubble--primary .bubble__content, .bubble_form input[type=radio]:not(old):checked + .bubble .bubble__content, .bubble--primary .bubble_form input[type=number]:not(old), .bubble_form .bubble--primary input[type=number]:not(old), .bubble_form input[type=radio]:not(old):checked + .bubble input[type=number]:not(old) {
```
msecret, line 712: Note to self: pay attention to generated css.

cmc333333:
I agree about the JS solution, unless there's a good way to implement parent selectors in CSS.

Some other todos:
- Add tabindexes or otherwise to aid accessibility
- Remove the second form from the python code

PR #95, opened 2014-11-13
api and JS to update project funding percentages automatically.

annalee:
paired on the JS with my buddy @zackferraro.


peacecorps/peacecorps/api.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+from restless.views import Endpoint
+
+from peacecorps.models import Account
+
+class GetAccountPercent(Endpoint):
+    def get(self, request):
+        code = request.params.get('code')
+        percent = Account.objects.get(code=code).percent_funded()
+        return {'percent': '%s' % percent}
```
cmc333333: Shouldn't this be a float or int rather than a string?

peacecorps/peacecorps/models.py
```diff
@@ -58,7 +58,10 @@ def total(self):
             return self.current
 
     def percent_funded(self):
-        return percentfunded(self.total(), self.goal)
+        if self.goal:
+            return percentfunded(self.total(), self.goal)
+        else:
+            return ''
```
cmc333333: This is returning a float in one case and a string in the other. Maybe it should just be `0` if no goal?

peacecorps/peacecorps/api.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+from restless.views import Endpoint
+
+from peacecorps.models import Account
+
+class GetAccountPercent(Endpoint):
+    def get(self, request):
+        code = request.params.get('code')
```
msecret: Based on my somewhat limited REST knowledge, I would've expected code to be in the url rather then a param. Something like this:

```
GET /api/accountpercent/{code}
```

Is there a reason it's in a GET param?

peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
```diff
@@ -0,0 +1,37 @@
...
+  'use strict';
+
+  PC.UpdatePercent = function(root){
+    this.root = root;
+  };
+  PC.UpdatePercent.prototype.getTotal = function(){
+    $.ajax({
+        url: '/api/accountpercent?code=' + this.code,
+        method: 'GET'
+    }).done(this.updateHTML.bind(this));
```
msecret, line 15: Good job using the Promises API!

peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
```diff
@@ -0,0 +1,37 @@
+/* Update donations percentages on project pages. */
+
+var PC = PC || {};
+
+(function($) {
+  'use strict';
+
+  PC.UpdatePercent = function(root){
+    this.root = root;
```
msecret: A common JS style guidelines is to name any jqueryized elements with a $ sign like this:

this.$root = $root.

peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
```diff
@@ -0,0 +1,37 @@
...
+    var text = this.root.find(".funded-amount-text");
+    bar.css('max-width', oData.percent+'%');
+    text.text(parseInt(
+              Math.round(oData.percent), 10) + '% funded');
+  };
+
+  PC.UpdatePercent.prototype.init = function(){
+    this.getCode();
+    var seconds = 60;
+    this.interval = setInterval(this.getTotal.bind(this), seconds * 1000);
```
msecret, line 31: This is fine for now, but we may want to suggest polling as the choice here to update. Polling is very bad for mobile devices in that it will kill the battery. I think we should discuss further down the road though.

peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  var updatePercent = new PC.UpdatePercent($('#fundingbar'));
```
msecret: I would suggest using a class rather then an ID here and prefixing the class with "js" so its known thats the purpose of it. So like this:

.js-fundingBar

peacecorps/peacecorps/templates/donations/includes/project-top.jinja
```diff
@@ -43,7 +43,9 @@
           </div>
             <span>in <strong>{{project.country.name}}</strong></span>
         </section>
-        <section class="showcase--col3_centered__col">
+        <section class="showcase--col3_centered__col"
+                 data-project-code="{{ project.account.code }}"
+                 id="fundingbar">
```
msecret: I would suggest adding this to class with js-fundingBar as mentioned above.

peacecorps/peacecorps/urls.py
```diff
@@ -75,7 +76,8 @@
         name='donation success'),
     url(r'^failure/?$',
         TemplateView.as_view(template_name='donations/failure.jinja'),
-        name='donation failure')
+        name='donation failure'),
+    url(r'^api/accountpercent$', api.GetAccountPercent.as_view()),
```
cmc333333: Would `api/account` make more sense? It'd allow us to expand the endpoint in the future.

`api/account/XXXX` would fit @msecret's suggestion above

cmc333333: Thinking about this a bit more, I don't know that `Accounts` are relevant to end users. When we start re-implementing the views to be API-driven, I think we'd only care about `Projects`. That leads me to think that the percentage should be within a `project` endpoint, but that sort of rewrite is probably out of scope for this pull request.

msecret:
Everything looks good to me, not sure if @cmc333333 wanted to take another look before merge.

cmc333333:
+1 thanks @annalee

PR #96, opened 2014-11-14
ehnc Finish project layout

msecret:
Added new bar to the project layout as specified in mock

TODO: unsure what to set the width of the bar to.
TODO: check responsiveness 

-- 2nd commit
Fixes to responsive layout
The responsive layout was broken at smaller breakpoints due to
using a max width. This new setup allows user to set positions
using straight percentages in markup and the css will take care of
rest.

@cmc333333 I haven't addressed your issues yet.

@ericronne I would like some help on how to deal with the layout at the smallest breakpoint. Here is a screenshot of the issues:
screenshot from 2014-11-14 09 13 05


peacecorps/peacecorps/templates/donations/donate_project.jinja
```diff
@@ -14,8 +14,52 @@
 <div class="u-clearfix"></div>
 
 <div class="showcase--main showcase--neutral">
-  <div class="showcase_inset--lg">
+  <div class="showcase_inset--lg bm3">
     <h5 class="t-heading_sm">give to {{project.title}}</h5>
+    {% set percent_completed = project.account.current / project.account.goal *100
+    %}
+    {% set percent_community = project.account.community_contribution / project.account.goal * 100
+    %}
```
cmc333333: Why not use `project.account.percent_funded` and add a similar method for community contribution? Putting this in the template leaves room for a lot of error -- the template will explode, for example, if `project.account.goal` is zero.

peacecorps/peacecorps/templates/donations/donate_project.jinja
```diff
@@ -14,8 +14,52 @@
...
+    <div class="rectangular_progress rectangular_progress--secondary">
+      <span class="rectangular_progress__point"
+            style="left:{{percent_community}}%;">
+        community constributions<br>
+        <strong>
+          ${{ "{:,}".format(project.account.community_contribution) }}
```
cmc333333: Use `humanize_amount` (you can pass it into the template context). The number you're showing here will be 100x larger than you intend.

peacecorps/peacecorps/templates/donations/donate_project.jinja
cmc333333: Also, I think we determined that the data underlying the "goal" does _not_ include the community contribution. I think that means that you want your denominator to be `project.account.goal + project.account.community_contribution`

cmc333333:
@annalee @msecret : be sure to sync up regarding this template as you have significantly conflicting pull requests

peacecorps/peacecorps/static/peacecorps/css/donation-form.css
```diff
@@ -1,91 +1,104 @@
...
- * MIT License */
-html {
-  -webkit-box-sizing: border-box;
-  -moz-box-sizing: border-box;
-  box-sizing: border-box; }
+/*
+Syntax error: Invalid CSS after "      webkit-image": expected ")", was ": -webkit- + $p..."
+        on line 21 of sass/bourbon/helpers/_linear-angle-parser.scss
+        from line 10 of sass/bourbon/_bourbon.scss
+        from line 1 of sass/donation-form.scss
```
cmc333333: Syntax error

peacecorps/peacecorps/templates/donations/donate_project.jinja
```diff
@@ -14,8 +14,49 @@
 <div class="u-clearfix"></div>
 
 <div class="showcase--main showcase--neutral">
-  <div class="showcase_inset--lg">
+  <div class="showcase_inset--lg bm3">
     <h5 class="t-heading_sm">give to {{project.title}}</h5>
+    {% set percent_completed = project.account.percent_funded() %}
```
cmc333333, line 19: This still isn't quite right. Currently, `percent_funded` is the money raised over the goal, and `percent_community_funded` is the community contribution over the goal + the community contribution. Those percents, then, aren't really comparable.

If we're assuming that when we want a percent, we always want a fraction of the _total_ amount of the project (i.e. the fund-raising goal + the community contribution), `percent_funded` should be modified to also be out of that sum.

Unrelated: why are you using a template variable for one of these percents and a method call for the other?

ericronne:
@msecret Here's a slightly tweaked version of the status-bar layout, along with mobile. Let me know if this doesn't cut it, or you have thoughts/q's. thx!
fundingstatus-11-16-14

onezerojeremy:
Can we please exchange the "donate the full amount" button with a regular donate button.

peacecorps/peacecorps/templates/donations/donate_project.jinja
msecret, line 19: No major reason, just that I had it set up similarly before and it was less to write. I think I might use a template var for both.

cmc333333:
Thanks @msecret ! We can address @onezerojeremy 's concerns with the next design update.

PR #99, opened 2014-11-17
Updates to import script

cmc333333:
Now that we've seen an example data file, the import script has been updated to account for issues found. I apologize for the size of the pull request -- it accounts for a lot of small issues.
- For the time being, accept datetimes in `%d-%b-%y` until the data can be cleaned up
- Previously, we had a manual mapping of "sector" values to campaign names. This takes a different strategy; whenever a sector fund is imported, an entry is added to the SectorMapping table. Determining the campaign associated with a given project using the "sector" field, then, is simply a lookup in that table
- `create_account_pcpp` has been split into a function for creating the account and campaign and a function for creation the project. Previously, we assumed that the list would only contain projects, so the logic needed to be altered to accommodate country, memorial, sector, and general funds.
- Added code to account for non-unique account names (the fund code is concatenated)
- Added code to account for non-unique project slugs (an integer is tacked on)
- Campaigns and Projects now automatically set their slug when saving.
- Multiple expected field names (as seen in the CSV) have been updated
- Description/summary is now imported
- The CSV does not include the current fund amount, but does include the fund's goal and "balance", so the logic is now a subtraction
- Determine if we're creating a project, country fund, memorial fund, etc. through heuristics (which match the provided data). See `account_type()`
- Assume a non-utf8 encoding to the import file

### DIFF
--- a/peacecorps/peacecorps/management/commands/sync_accounting.py
+++ b/peacecorps/peacecorps/management/commands/sync_accounting.py
@@ -14,8 +14,12 @@ def datetime_from(text):
     """Convert a string representation of a datetime into a UTC datetime
     object."""
     # This format will no doubt change
-    return timezone.make_aware(datetime.strptime(text, "%Y-%m-%d %H:%M:%S"),
-                               timezone.get_current_timezone())
+    try:
+        time = datetime.strptime(text, "%Y-%m-%d %H:%M:%S")
+    except ValueError:
+        # Try a different format. @todo: remove once data's cleaned
+        time = datetime.strptime(text, "%d-%b-%y")
+    return timezone.make_aware(time, timezone.get_current_timezone())
 
 
 def cents_from(text):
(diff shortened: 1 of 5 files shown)


peacecorps/peacecorps/management/commands/sync_accounting.py
```diff
@@ -26,49 +31,100 @@ def cents_from(text):
...
+            name=name, account=account, campaigntype=acc_type,
+            description=row['SUMMARY'])
+        if acc_type == Account.SECTOR:
+            # Make sure we remember the sector this is marked as
+            SectorMapping.objects.create(pk=row['SECTOR'], campaign=campaign)
+
+
+def create_pcpp(account, row, issue_map):
+    """Create and save a project (and account). This is a bit more complex for
+    projects, which have foal amounts, etc."""
```
msecret, line 73: foal = goal?

peacecorps/peacecorps/management/commands/sync_accounting.py
```diff
@@ -80,12 +136,14 @@ def handle(self, *args, **kwargs):
         if len(args) == 0:
             raise CommandError("Missing path to csv")
 
-        with open(args[0]) as csvfile:
+        issue_map = IssueCache()
+
+        with open(args[0], encoding='iso-8859-1') as csvfile:
```
msecret, line 141: Would setting an encoding here potential make it so csv files from certain programs or OSs not work?

PR #101, opened 2014-11-18
Jshint, callback immediately, 60 seconds

cmc333333:
- JSHint fixes
- Make the callback immediately, in case the user is seeing a cached page
- Change from updating every 5 seconds to every 60

### DIFF
--- a/peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
+++ b/peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
@@ -1,9 +1,9 @@
 /* Update donations percentages on project pages. */
+'use strict';
 
 var PC = PC || {};
 
 (function($) {
-  'use strict';
 
   PC.UpdatePercent = function(root){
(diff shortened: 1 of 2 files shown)


peacecorps/peacecorps/static/peacecorps/js/update_donatepercent.js
```diff
@@ -15,11 +15,11 @@ var PC = PC || {};
     }).done(this.updateHTML.bind(this));
   };
   PC.UpdatePercent.prototype.getCode = function(){
-    this.code = this.$root.data("project-code");
+    this.code = this.$root.data('project-code');
   };
   PC.UpdatePercent.prototype.updateHTML = function(oData){
-    var bar = this.$root.find(".funded-amount-bar");
-    var text = this.$root.find(".funded-amount-text");
+    var bar = this.$root.find('.funded-amount-bar');
```
msecret, line 21: Cool thanks for being precise like this.

PR #102, opened 2014-11-18
Some logging events

cmc333333:
- When pay.gov makes a callback, log the success or failure
- Make sure to delete donor info when above callback is a failure
- When clearing stale donor information, log how many were removed
- When synchronizing account information, log when a new account is added, updated, or a failure occurred (due to invalid country/sector)
- Remove the reliance on the 'tests.yaml' fixture from the synchronization tests. Locally that sped test execution from 35s to 22s

### DIFF
--- a/peacecorps/paygov/views.py
+++ b/peacecorps/paygov/views.py
@@ -40,7 +40,7 @@ def results(request):
     # Transaction was canceled or had another error
     elif request.POST.get('payment_status') != 'Completed':
         message = request.POST.get('error_message', 'Unknown error')
-        logger.info("Transaction %s: %s", request.POST.get('payment_status'),
+        logger.warn("Transaction %s: %s", request.POST.get('payment_status'),
                     message)
         info.delete()
     elif not request.POST.get('payment_amount'):
--- a/peacecorps/paygov/views.py
+++ b/peacecorps/paygov/views.py
@@ -30,27 +30,31 @@ def results(request):
     message = 'OK'
     if request.method != 'POST':
         return HttpResponseNotAllowed(['POST'])
-    if not request.POST.get('agency_tracking_id'):
-        message = 'Missing agency_tracking_id'
+
+    info = DonorInfo.objects.select_related('account').filter(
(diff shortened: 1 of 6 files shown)


peacecorps/peacecorps/management/commands/sync_accounting.py
```diff
@@ -137,6 +138,7 @@ def handle(self, *args, **kwargs):
             raise CommandError("Missing path to csv")
 
         issue_map = IssueCache()
+        logger = logging.getLogger('peacecorps.sync_accounting')
```
msecret, line 141: One thing we did at my last company that might help is get the logger as a var at the top of the file. I'll leave it to you if you'd feel like that would help.

``` python
#imports 

_log = logging.getLogger

def datetime_from():
...
```

peacecorps/paygov/views.py
```diff
@@ -30,27 +30,31 @@ def results(request):
...
+    info = DonorInfo.objects.select_related('account').filter(
+        pk=request.POST.get('agency_tracking_id')).first()
+    if not info:
+        message = 'Invalid agency_tracking_id'
     elif not request.POST.get('payment_status'):
         message = 'Missing payment_status'
+    # Transaction was canceled or had another error
     elif request.POST.get('payment_status') != 'Completed':
         message = request.POST.get('error_message', 'Unknown error')
+        logger.info("Transaction %s: %s", request.POST.get('payment_status'),
```
msecret: It seems like something went wrong in this case, would we want to log.error or log.warn?

peacecorps/peacecorps/management/commands/sync_accounting.py
cmc333333, line 141: I'm not a fan of global state :)

peacecorps/paygov/views.py
cmc333333: That's a good question -- I'm not too sure how common canceled transactions are. Let's bet on uncommon; I'll change this to a warning.

PR #108, opened 2014-11-21
Refc typography spacing

msecret:
Basic idea of spacing for typography.

I found that I was importing completely incorrectly in sass (even
though importing works in this way in most programming languages). I
should only import each file once unless its vars in mixins, otherwise
it will show up multiple times in the output. This means mixins all
have to be in one file.


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -10,38 +10,18 @@
...
-                </li>
-                <li class="main_nav__link">
-                  <a href="/volunteer">Volunteer</a>
-                </li>
-                <li class="main_nav__link">
-                  <a href="{{ url("donate landing") }}">Donate</a>
-                </li>
-            </ul>
-        </nav>
-    </header>
```
cmc333333, line 34: Just verifying, you want all this to go away and not be replaced with anything, yes?

peacecorps/peacecorps/static/peacecorps/sass/_mixins.scss
```diff
@@ -6,3 +6,48 @@
...
+
+@mixin u-visually_hidden {
+  position: absolute !important;
+  top: -9999px !important;
+  left: -9999px !important;
+}
+
+@function parseInt($n) { /* 2 */
+  @return $n / ($n * 0 + 1);
+}
```
cmc333333, line 24: Would you mind commenting this? I'm guessing it takes a string and converts it to an int based on its name, but I don't understand they multiplication by zero at all.

Why are there integers next to many of these functions? Depth level?

peacecorps/peacecorps/static/peacecorps/sass/_mixins.scss
```diff
@@ -6,3 +6,48 @@
...
+
+@mixin rem($property, $values) {
```
cmc333333, line 34: Similar here: would you mind commenting how to use this function and what it does?

peacecorps/peacecorps/templates/donations/base.jinja
msecret, line 34: Yea we're removing the nav in this mockup. It may come back at some point, but I'm fine removing it for now.

cmc333333:
thanks!

PR #112, opened 2014-11-26
ehnc Backend for sorter

msecret:
I'd like some advice on the following:
- How I passed a []Volunteer object down. I think theres a better way
  to do this, and some nice python way of copy those properties over
- My naming of the new routes. I referred to REST principals of creating
  a new resource for this as one doens't really exist for it. Is "all" ok?

TODO:
- I found an issue where the multiple <object> tags are not always using
  the cache to get the svg image. This is causing performance problems and
  has to be fixed.

### DIFF
--- a/peacecorps/peacecorps/templates/donations/donate_all.jinja
+++ b/peacecorps/peacecorps/templates/donations/donate_all.jinja
@@ -67,7 +67,6 @@
               <div class="rectangular_progress rectangular_progress--secondary">
                 {# TODO make macro #}
                 {% set percent_completed = project.account.percent_funded() %}
-                {% set percent_community = project.account.percent_community_funded() %}
                 <div class="rectangular_progress__bar">
                   <div class="rectangular_progress__filled"
                       style="width: {{ percent_completed }}%;">
@@ -77,7 +76,8 @@
 
               <a class="button button--sm button--primary" href="">
                Give to the project</a>
-              <a href="/donate/project/{{ project.slug }}">Read more</a>
+              <a href="{{ url("donate project", slug=project.slug) }}">
+                Read more</a>
             </td>
           </tr>
         {% endfor %}
--- a/peacecorps/peacecorps/templates/donations/donate_all.jinja
+++ b/peacecorps/peacecorps/templates/donations/donate_all.jinja
@@ -1,0 +1,106 @@
+{% extends "donations/base.jinja" %}
+
+{% block title %}PEACECORPS{% endblock %}
+{% block header %}PEACECORPS{% endblock %}
+{% block content_class %}landing{% endblock %}
+
+{% block content %}
+
+<table>
+  <tr>
+    <td>
+      <table name="issues">
+        {% for issue in issues %}
+          <tr name="{{ issue.name }}">
+            <td><img width=50" src="{{ issue.icon.url }}"
+                     alt="{{ issue.name }}" /></td>
+            <td><h3>{{ issue.name }}</h3></td>
+          </tr>
+        {% endfor %}
+      </table>
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <table name="countries">
+        {% for country in countries %}
+          <tr name="{{ country.name }}">
+            <td>
+              {{ country_map(country.code) }}
+            </td>
+            <td><h3>{{ country.name }}</h3></td>
+          </tr>
+        {% endfor %}
+      </table>
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <table name="volunteers">
+        {% for volunteer in volunteers %}
+          <tr name="{{ volunteer.name }}">
+            <td>
+              <img src="{{ volunteer.picture.url }}"
+                   alt="{{ volunteer.name }}" />
+            </td>
+            <td>
+              <h3>{{ volunteer.name }}</h3>
+              <p>{{ volunteer.homestate }}, {{ volunteer.homecountry }}</p>
+            </td>
+          </tr>
+        {% endfor %}
+      </table>
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <table name="projects">
+        {% for project in  projects %}
+          <tr name="{{ project.title }}">
+            <td>
+              <img src="{{ volunteerpicture }}" alt="{{ volunteername }}" />
+            </td>
+            <td>
+              <h3>{{ project.title }}</h3>
+              <p>{{ project.description }}</p>
+
+              <div class="rectangular_progress rectangular_progress--secondary">
+                {# TODO make macro #}
+                {% set percent_completed = project.account.percent_funded() %}
+                {% set percent_community = project.account.percent_community_funded() %}
+                <div class="rectangular_progress__bar">
+                  <div class="rectangular_progress__filled"
+                      style="width: {{ percent_completed }}%;">
+                  </div>
+                </div>
+              </div>
+
+              <a class="button button--sm button--primary" href="">
+               Give to the project</a>
+              <a href="/donate/project/{{ project.slug }}">Read more</a>
+            </td>
+          </tr>
+        {% endfor %}
+      </table>
+    </td>
+  </tr>
+</table>
+
+{% endblock %}
+
+{% block custom_js %}
+<script type="text/javascript" src="{{ static("peacecorps/js/country_map.js")
+    }}"></script>
+{% endblock %}
+
+{% macro country_map(code) -%}
+  <div class="mask mask--oval mask--bordered u-h_center--block map">
+    <object data-country="{{code}}"
+            class="mask__content js-worldmap"
+            id="js-worldmap"
+            type="image/svg+xml"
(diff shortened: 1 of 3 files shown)


peacecorps/peacecorps/urls.py
```diff
@@ -13,10 +13,13 @@
     '',
     url(r'^donate/?$', midterm_cache(views.donate_landing),
         name='donate landing'),
+    url(r'^donate/all/?$', midterm_cache(views.donate_all), name='donate all'),
 
     url(r'^donate/campaign/' + _slug + r'/?$',
         midterm_cache(views.donate_campaign), name='donate campaign'),
     # not cached so the values are up-to-date
+    url(r'^donate/campaign/' + _slug + r'/?$', views.donate_campaign,
+        name='donate campaign'),
```
cmc333333: This duplicates the above

peacecorps/peacecorps/views.py
```diff
@@ -220,6 +220,35 @@ def donate_general(request, slug):
...
+                campaigntype=Campaign.SECTOR).order_by('name')
+    projects = Project.published_objects.select_related('country', 'account')
+    volunteers = []
+    for project in projects:
+        volunteer = Volunteer()
+        volunteer.name = project.volunteername
+        volunteer.picture = project.volunteerpicture
+        volunteer.homestate = project.volunteerhomestate
+        volunteer.homecity = project.volunteerhomecity
+        volunteers.append(volunteer)
```
cmc333333: Using a dict may make more sense as the class isn't used elsewhere. The template won't need to change (dot notation works as dictionary key access within jinja).

``` python
volunteers.append({
    'name': project.volunteername,
    'picture': project.volunteerpicture,
    'homestate': project.volunteerhomestate,
    'homecity': project.volunteerhomecity
})
```

or, perhaps more cryptically:

``` python
volunteers.append({key: getattr(project, 'volunteer' + key)
                   for key in ('name', 'picture', 'homestate', 'homecity')})
```

peacecorps/peacecorps/views.py
```diff
@@ -220,6 +220,35 @@ def donate_general(request, slug):
...
         })
 
+def donate_all(request):
+    """
+    The page that displays a sorter for all projects, issues, volunteers.
+    """
+    class Volunteer():
+        pass
+
+    countries = Campaign.objects.all()
```
cmc333333: Should this be `Campaign.objects.filter(campaigntype=Campaign.COUNTRY)`?

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -0,0 +1,106 @@
...
+              <img src="{{ volunteer.picture.url }}"
+                   alt="{{ volunteer.name }}" />
+            </td>
+            <td>
+              <h3>{{ volunteer.name }}</h3>
+              <p>{{ volunteer.homestate }}, {{ volunteer.homecountry }}</p>
+            </td>
+          </tr>
+        {% endfor %}
+      </table>
```
cmc333333, line 52: Something to consider: as there's an assumed one-to-one mapping between volunteers and projects, there may not be a need for a separate volunteer table.

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -0,0 +1,106 @@
...
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <table name="countries">
+        {% for country in countries %}
+          <tr name="{{ country.name }}">
+            <td>
+              {{ country_map(country.code) }}
```
cmc333333, line 29: Based on the view's code, `countries` is a collection of campaigns, which don't have a country code. We'll need to think this through a bit more carefully -- I'm surprised the template doesn't crash. I'm cool merging and fixing later so long as this isn't currently blowing up.

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -0,0 +1,106 @@
...
+              <img src="{{ volunteerpicture }}" alt="{{ volunteername }}" />
+            </td>
+            <td>
+              <h3>{{ project.title }}</h3>
+              <p>{{ project.description }}</p>
+
+              <div class="rectangular_progress rectangular_progress--secondary">
+                {# TODO make macro #}
+                {% set percent_completed = project.account.percent_funded() %}
+                {% set percent_community = project.account.percent_community_funded() %}
```
cmc333333: Looks like this variable isn't used

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -0,0 +1,106 @@
...
+                <div class="rectangular_progress__bar">
+                  <div class="rectangular_progress__filled"
+                      style="width: {{ percent_completed }}%;">
+                  </div>
+                </div>
+              </div>
+
+              <a class="button button--sm button--primary" href="">
+               Give to the project</a>
+              <a href="/donate/project/{{ project.slug }}">Read more</a>
```
cmc333333: `{{url("donate project", slug=project.slug)}}` -- this indirection us to change the url later.

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -0,0 +1,106 @@
...
+{% endblock %}
+
+{% macro country_map(code) -%}
+  <div class="mask mask--oval mask--bordered u-h_center--block map">
+    <object data-country="{{code}}"
+            class="mask__content js-worldmap"
+            id="js-worldmap"
+            type="image/svg+xml"
+            data="{{ static("peacecorps/img/blankmap-world6.svg") }}">
+    </object>
```
cmc333333, line 103: I wonder, is there any way to have a single svg with multiple views on it? I worry that we'll overload mobile devices. Maybe we pregenerate all of the images and attach them to the country model.

cmc333333:
I know @ericronne had some reservations about "all". To be pedantic, this page doesn't list all funds -- the memorial funds aren't here, nor is the general fund (not to be confused with the global fund).

Regarding code - there's a few minor things + the concern over using `country.code`. This is a good start, though. @annalee we probably need to chat again about models -- now "issues" may have more than one associated sector, meaning the "campaign" strategy needs some tuning.

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret, line 52: Thats true, although I may need this structure when I'm writing the JS. Maybe I'll keep it in for now until I can determine if I need it or not.

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret: Yea copy pasted it. I'm going to potentially make a macro for the whole bar at some point.

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret, line 103: Yea I'm going to look at a solution to this. I like both those ideas though.

msecret, line 103: The only thing with pre-generating images: it might be good to keep them as svgs so we can scale them easily in different ways. That would still be possible if we pre-generated though, so that might be the best idea.

peacecorps/peacecorps/urls.py
msecret: Sorry merge conflict.

peacecorps/peacecorps/views.py
msecret: Thats so cool dot notion works on dicts in jinja! That was one of the larger reasons I did this.

Also, thanks for sharing some

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret, line 29: I think I had meant to use the Country model, which is part of the UI.

Although there are other things here: I will also need to add the country fund for each country. Whats the relation between Campaign of type country -> country.name or country.code?

cmc333333, line 29: That's a great question, I'll create an issue. Starting with `Country.objects.all` should be okay until that's sorted out.

peacecorps/peacecorps/urls.py
```diff
@@ -13,10 +13,11 @@
     '',
     url(r'^donate/?$', midterm_cache(views.donate_landing),
         name='donate landing'),
+    url(r'^donate/all/?$', midterm_cache(views.donate_all), name='donate all'),
 
-    url(r'^donate/campaign/' + _slug + r'/?$',
-        midterm_cache(views.donate_campaign), name='donate campaign'),
     # not cached so the values are up-to-date
+    url(r'^donate/campaign/' + _slug + r'/?$', views.donate_campaign,
```
cmc333333: Minor: this was wrapped in a midterm_cache before

PR #113, opened 2014-11-26
Use S3 for hosting of static files

cmc333333:
- Create a mechanism for separating static and media files. The built in S3boto storage assumes a single bucket for both static assets (theme images, js, css, etc.) and media (image uploads). This commit uses different classes to get around that assumption, which could lead to clobbered files.
- Moved the inclusion of "storages" from all web servers to admin web servers -- there's no need to upload from anywhere else
- Renamed buckets we are using. These have been prepopulated with the appropriate files
- Configured all web servers to use the s3 bucket when serving static assets. The deployment process runs `collectstatic` on an admin server, sending all the static files to s3.

### DIFF
--- a/peacecorps/contenteditor/backends.py
+++ b/peacecorps/contenteditor/backends.py
@@ -4,6 +4,8 @@ from django.conf import settings
 from django.contrib.auth import backends
 from django.utils.functional import empty, LazyObject
 from django.utils.module_loading import import_string
+# Would it be better to move this requirement elsewhere?
+from storages.backends.s3boto import S3BotoStorage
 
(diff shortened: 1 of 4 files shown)


peacecorps/contenteditor/backends.py
```diff
@@ -4,6 +4,8 @@
 from django.contrib.auth import backends
 from django.utils.functional import empty, LazyObject
 from django.utils.module_loading import import_string
+# Would it be better to move this requirement elsewhere?
```
msecret, line 7: This seems fine to me

PR #114, opened 2014-12-01
Load SVG map via ajax (fixing CORS errors)

cmc333333:
- Moves svg's css into the file, reducing issues with using a relative path (and reducing the number of HTTP requests)
- Fixes typo in jshintrc
- Uses ajax to load svg files (instead of the object tag). This allows us to add the 'Origin' header, needed to account for the cross-domain image manipulation

Note that this does nothing to fix the too-many-images-on-one-page issue and most likely exacerbates it, since the same 1.5 meg is being inserted many times on the same page.


peacecorps/peacecorps/static/peacecorps/js/country_map.js
```diff
@@ -122,20 +122,26 @@ var PC = PC || {};
...
-
-    elMap = document.getElementById('js-worldMap');
-    svg = elMap.contentDocument.querySelectorAll('svg')[0];
-    selectedCountryCode = elMap.getAttribute('data-country');
-
-    countryMap = new PC.CountryMap(svg, selectedCountryCode);
-    countryMap.init();
+  // main - if necessary, fetch the map and update countries
+  $(document).ready(function() {
+    var $mapEls = $('.map');
```
msecret, line 127: This selector should be something like .js-map

peacecorps/peacecorps/static/peacecorps/js/country_map.js
```diff
@@ -122,20 +122,26 @@ var PC = PC || {};
...
+        $mapEls.each(function(idx, mapEl) {
+          var $mapEl = $(mapEl),
+              countryMap;
+          $mapEl.append(svgEl);
+          countryMap = new PC.CountryMap($mapEl.children()[0],
+                                         $mapEl.data('country'));
+          countryMap.init();
+        });
+      });
+    }
```
msecret, line 145: So if we know they are all going to be the same URL for the map svg, we could load it once from ajax and append it for each of them, correct?

Wouldn't this potentially fix the issue of having to load such a large number of svg images?

peacecorps/peacecorps/static/peacecorps/js/country_map.js
cmc333333, line 127: Can do

peacecorps/peacecorps/static/peacecorps/js/country_map.js
cmc333333, line 145: This limits the bandwidth used as the image is only downloaded once, but blows up the DOM (as the 1.5M file is effectively being inserted dozens of times)

PR #115, opened 2014-12-01
Moving Campaign and Project descriptions to Sir Trevor

annalee:
Switches the description fields on Projects and Campaigns to use Sir Trevor instead of TinyMCE.

Sir Trevor stores markdown inside json, which isn't playing nicely with manage.py's dumpdata command. I've manually edited the test fixture while I figure out a better fix for the problem. Running the included data migration will get  the current test data working.

Still need:
[ ] CSS fix to keep the new fields from doing ugly things to the layout
[ ] image resizing
[ ] 508-compliant image blocks for Sir Trevor
[ ] an image library that works with Sir Trevor, if we can get it


peacecorps/peacecorps/templatetags/markdown.py
```diff
@@ -0,0 +1,6 @@
+from markdown import markdown
+from django_jinja import library
+
+@library.filter
+def markdown(text):
+    return markdown(text)
```
cmc333333: Is this file still needed? I think the `.html` stuff doesn't use jinja

peacecorps/peacecorps/urls.py
```diff
@@ -11,6 +11,7 @@
 
 urlpatterns = patterns(
     '',
+    url(r'^sirtrevor/', include('sirtrevor.urls')),
```
cmc333333: Is this only needed for the admin? If so, can we conditionally add it?

cmc333333:
Nice work on the migrations. I think you may have missed an update to `requirements.txt`. Does anything need to be added to `INSTALLED_APPS`? Can we remove all references to tinymce?

peacecorps/peacecorps/templatetags/markdown.py
annalee: ah, good call. I'll remove.

cmc333333:
Looks good (but yay automated tests!)

PR #116, opened 2014-12-02
Separate SVG country files

cmc333333:
`cropmap.py` implements `country_map.js`, but as a one-off script that generated many individual country files. We can tweak this script as needs arise, and re-run to generate new svgs.

The script strips out country data outside the view box, but it could certainly be modified to simplify the shapes to reduce file size.

The script requires an external library, but as that library isn't needed for normal operations, and doesn't have a `setup.py`, I've just added instructions for how to get it.


cmc333333:
Due to the new svgs, github gave up trying to render a diff. Here's the relevant diff: https://gist.github.com/cmc333333/e83a36e2accf08d75b61

PR #117, opened 2014-12-03
Add field for country to campaign. Use that during synchronization

cmc333333:
We need a way to grab country funds' associated country picture. Instead of using an ugly `LIKE` query, this explicitly stores the relevant country on the campaign object.

For now, it's unique (as only country funds use it), but we could remove that restriction if the country associated with memorial funds becomes relevant.

You'll notice some migrations to other fields; those are just description edits that weren't in `master`'s migrations.

### DIFF
--- a/peacecorps/peacecorps/management/commands/sync_accounting.py
+++ b/peacecorps/peacecorps/management/commands/sync_accounting.py
@@ -58,19 +58,35 @@ def create_account(row, issue_map):
     account = Account(name=name, code=row['PROJ_NO'], category=acc_type)
     if acc_type == Account.PROJECT:
         create_pcpp(account, row, issue_map)
-    else:   # Campaign
-        account.save()
-        campaign = Campaign.objects.create(
-            name=name, account=account, campaigntype=acc_type,
-            description=row['SUMMARY'])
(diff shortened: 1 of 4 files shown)


msecret:
So this will mean when working with campaign objects, there'll be conditionals to check if its of type country?

if campaign.campaigntype == country:
  do_something_with(campaign.country)

I think that would be fine, I just wanted to get an idea of how you're envisioning this.

cmc333333:
Currently, on the filter/table/sorter page, a list of all countries is provided to the template. This isn't quite accurate, as not every country has a fund and jumping from country to associated fund isn't easy. With this patch, we'd instead pass in a list of country funds (i.e. `Campaign.objects.filter(campaigntype=Campaign.COUNTRY)` or similar). We'd have access to the country code (needed for the graphic) as well as the fund information (needed to direct donations).

Generally, we'll know which sub-type of Campaign (memorial fund, country fund, sector, etc.) we'll need and filter accordingly. We shouldn't need to check the type in templates too often, then. It does make me wonder if we should encourage the division of fund type further via proxy models for MemorialFund, CountryFund, etc., though.

annalee:
Looks good. Merging.

PR #118, opened 2014-12-03
Build setup

msecret:
Here's a first pass at creating a JS build. This is a work in progress,
please DONT MERGE it. I'm looking for early feedback.

I used grunt for the following reasons:
- Because its nodejs, it works well on Windows.
- Its easier to use and understand for node newbies then other tools
  such as Gulp.
- There are lots of plugins for it, which *can be easier for node
  newbies
- if the plugins actually work.

I used browserify:
- Because its cool.

This build's purpose:
- Keep code consistency by linting.
- Provide framework for unit, integration testing of frontend code.
- Build JS, CSS, image assets.
- Track and enforce certain performance metrics.
- Organize front end files in one place.
- Bundle 3rd party dependencies.
- Provide an easy way for developers to download and work on the
  project.

TODO:
- Unit tests, close to finished.
- Figure out best way to ensure JS code that runs on specific pages
  doesn't run elsewhere.
- Add CSS to this build rather the separate. This will be done later
  down the road.

Other changes:
- Created a README file for the static dir as its becoming its own
  thing.


peacecorps/peacecorps/static/peacecorps/js/compiled/peacecorps-donation.min.js.map
```diff
@@ -0,0 +1 @@
...AAA,oBACA,SAAA,EAAA,GACA,MAAA,GAGA,EAAA,KAAA,GAAA,QAAA,gBACA,GAAA,EAAA,gBAJA,SAUA,EAAA,MACA,OAAA,GACA,QAAA,GACA,OAAA,SACA,SAAA,EAAA,GACA,EAAA,SAAA,EAAA,IACA,OAAA,SAAA,GAOA,IANA,GAAA,GAAA,EACA,KAGA,EAAA,gBAAA,GAAA,EAAA,MAAA,MAAA,GAEA,EAAA,EAAA,IACA,EAAA,EAAA,GAAA,GAAA,GACA,EAAA,IAAA,EAAA,EAAA,IAAA,EAAA,EAGA,OAAA,KAIA,GAAA,KAAA,KACA,EAAA,SAAA,EAAA,GAAA,IAAA,KAIA,EAAA,GAAA,QACA,IAAA,SAAA,EAAA,GACA,MAAA,IAAA,KAAA,SAAA,EAAA,EAAA,GACA,GAAA,GAAA,EACA,KACA,EAAA,CAEA,IAAA,EAAA,QAAA,GAAA,CAIA,IAHA,EAAA,GAAA,GACA,EAAA,EAAA,OAEA,EAAA,EAAA,IACA,EAAA,EAAA,IAAA,EAAA,IAAA,EAAA,EAAA,IAAA,EAAA,EAGA,OAAA,GAGA,MAAA,UAAA,EACA,EAAA,MAAA,EAAA,EAAA,GACA,EAAA,IAAA,EAAA,IACA,EAAA,EAAA,UAAA,OAAA,IAEA,KAAA,WACA,MAAA,GAAA,MAAA,IAEA,KAAA,WACA,MAAA,GAAA,OAEA,OAAA,SAAA,GACA,MAAA,iBAAA,GACA,EAAA,KAAA,OAAA,KAAA,OAGA,KAAA,KAAA,WACA,GAAA,MACA,EAAA,MAAA,OAEA,EAAA,MAAA,YAUA,EAAA,MAAA,EAEA,EAAA,WACA,YAAA,EACA,KAAA,SAAA,EAAA,EAAA,EAAA,EAAA,EAAA,GACA,KAAA,KAAA,EACA,KAAA,KAAA,EACA,KAAA,OAAA,GAAA,QACA,KAAA,QAAA,EACA,KAAA,MAAA,KAAA,IAAA,KAAA,MACA,KAAA,IAAA,EACA,KAAA,KAAA,IAAA,EAAA,UAAA,GAAA,GAAA;EAEA,IAAA,WACA,GAAA,GAAA,EAAA,UAAA,KAAA,KAEA,OAAA,IAAA,EAAA,IACA,EAAA,IAAA,MACA,EAAA,UAAA,SAAA,IAAA,OAEA,IAAA,SAAA,GACA,GAAA,GACA,E
```
cmc333333: Should this file be empty?

cmc333333: Nevermind, github just wasn't displaying.

cmc333333:
I like the browserify approach to JS requirements. I'm not convinced that will ever _need_ many additional libs, but this encourages more modular code and running grunt is no more onerous than sass. npm is a slightly different story, but we'll encounter that later, when we run tests more frequently.

Random thoughts:
- Everything in the static directory gets sent to S3. Clearly not a worry here, but something to keep in mind. Perhaps it makes more sense to run grunt outside of the static dir.
- We must be extra careful about licenses if we are packaging third party libs. Since we now distribute the files, we have more restrictions.
- I'm a bit antsy that all of the JS is included regardless of page. The extra overhead doesn't bother me so much as the lack of isolation. All initialization scripts need to account for being included on any page, which adds a subtle dependency on classes and ids throughout the site. Similarly, a small error in the JS relevant for one page would effectively shut down the site, as all pages would be broken. Of course, we should catch these in testing, but it makes me nervous
- Go ahead and add the npm installation and grunt test commands to circleci

msecret:
@cmc333333 I have a better idea for setting up folder structure, based on your first comment about the static directory.

I'm thinking it might be better to have a public directory where both compiled CSS, JS and images go, and have this put on the CDN rather then the whole static folder. Structure would be like this:

```
/static
Gruntfile, etc
  /public
    /js
    /css
    /img
  /src
    /js
    /css or sass
```

Let me know your thoughts

msecret:
Other responses:
- I'll check our license. I know I've done this at other companies and it was not a problem. Another thing: I'm eventually planning to bundle 3rd party libs in a separate JS file and including.
- I'm figuring out a solution to this currently, although I don't think it'll be too hard. I think at previous jobs it was solved by checking for the presence of a selector. Something to note: this bundle should change as the project changes. As we add more JS, we may find we want to split this up in different ways.
- Will do

cmc333333:
On public/private - we might be able to do that with symbolic links or custom rules for `collectstatic`, but it's probably not worth the effort right now.

On licenses, we just need to be thoughtful. The current [license](https://github.com/18F/peacecorps-site/blob/master/LICENSE.md) says that the code is public domain in the US and CC0 elsewhere, but that's not the case if we are distributing other libraries. CFPB provides a good [example](https://github.com/cfpb/regulations-site/blob/master/TERMS.md#exceptions)

On classes, most of the templates make use of the `content_class` jinja block, which may help.

LMK when you're comfortable with this being merged.

msecret:
On public/private, I'm going to keep the setup I have for now. I think it will be trivial to change around the dirs to have separate src and compiled dirs

On licenses, I've talked to Raphael about it, he's planning on getting back to me.

I came up with a solution for now for classes that will work.

msecret:
Also, @cmc333333, I'm ready for this to be merged.

rjmajma:
Hey @msecret. Just took a peek and @cmc333333 is absolutely right. Once you start using software that is already licensed, you must adjust your terms to specifically pass through the appropriate licenses. The link CM provides is a great way to articulate that (and something we should just steal as an org).

msecret:
Thanks @rjmajma and @cmc333333. Based on that example, it looks like I would only need to add regular dependencies, not development dependencies. This means just jQuery for now, and not the various grunt plugins and other.

Am I correct in thinking that?

konklone:
I think it's anything that appears _in_ the repo that isn't ours to release under CC0. Things listed as dependencies that are installed dynamically through a package manager (e.g. `pip`) wouldn't matter, but `.js` files that appear in the repo would, whether they're development dependencies or not.

cmc333333:
@msecret That sounds right to me.

My understanding is that we are complying with all licenses in production (i.e. when including the libraries in executable code, sending them to users, etc.). This is pretty easy because we're using open source libraries. The problem with the compiled JS is that we're _also_ distributing it on github, so we need to abide by the licenses' redistribution clauses. jQuery's under the MIT license, which only requires that the license be present in distributions.

PR #119, opened 2014-12-05
Ehnc project filter start js

msecret:
Started on JS for the filter UI. Added part that would show <tr>s based on what filter was selected.


peacecorps/peacecorps/static/peacecorps/js/src/discover.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+var $ = require('jquery');
+
+var Discover = function($root) {
+  // TODO move this all into shared code.
+  if ($root.length < 1) {
+    new Error('selector missing');
```
cmc333333: Should this be thrown/raised?

peacecorps/peacecorps/static/peacecorps/js/src/donation.js
```diff
@@ -77,4 +79,11 @@ $().ready(function() {
   Landing.createDataFilter();
 
   new UpdatePercent($('.js-fundingBar'));
+
+  if ($discoverApp) {
+    var discover = new Discover($('.js-discoverApp'), $('.js-discoverNav a'), {
```
cmc333333: You can reuse `$discoverApp` here, I think

cmc333333:
I like the approach of using attributes. It won't work for text search, but that's another challenge altogether.

cmc333333:
Thanks!

PR #120, opened 2014-12-05
Vignettes

cmc333333:
To move text/images out of the template, this creates a model called "Vignette", which is a unique piece of content with a Sir Trevor field. This will allow the content editors to modify the text, order, and elements present where they would otherwise have static content.

This branch also adds a "Button" block type for Sir Trevor, showing some of the potential benefits of that editor. Button blocks are styled exactly as we want them, and they're smart enough to try to float above immediately prior photos.

All in all, this is a tech demo, to see if this is a strategy we want to follow for donations and elsewhere on the site.
- Moved all admin info into the `contenteditor` app
- Added templates, classes, and scripts for the "Button" block
- Overrode the "heading" block template to fix an escaped ampersand
- Adds a fixture for some initial content for the two vignettes on the landing page. These reference s3 images, and will need to be replaced when @annalee's work on images is complete
- Vignette model, with a function to fetch a specific instance. If that instance isn't in the db, a warning is provided
- Moved `contenteditor` above `sirtrevor` so that it will override the "heading" template (and others if we choose)
- Added some demo styles. I'm not at all attached to these, just wanted to see things look not-awful
- Similarly, there's functionality in the Button block that keeps track of if the button should float above an image. This is just a tech demo; we shouldn't keep it for long
- PEP8


peacecorps/peacecorps/static/peacecorps/css/src/landing.scss
```diff
@@ -0,0 +1,19 @@
...
+  img {
+    width: 100%;
+    max-height: 350px;
+  }
+  .text-block {
+    @include span-columns(8);
+    @include shift(2);
+    text-align: center;
+  }
+}
```
msecret, line 14: Based on how the other CSS is written at this point a banner class that would go in its own banner.scss file (I'm also unsure of the name banner). From here both image and text block would be elements of the banner class. The css would look like this:

components/banner.scss

``` css
.banner {
  @include span-columns(12);
}
.banner__img {
  width: 100%;
  ...
}
.banner__text {
  @include span-columns(8);
  ...
}
```

peacecorps/peacecorps/static/peacecorps/css/src/landing.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.showcase--main.float-above {
```
msecret, line 16: A problem I find with this float-above class is that its using absolute numbers for positioning, so its very unusable. I think theres a way to do this so it can work with a div of any height. I can look at it.

peacecorps/peacecorps/static/peacecorps/css/src/landing.scss
msecret, line 14: I also don't want to have a "landing.scss" file, as I'm trying to move away from page specific styles, but towards a modular approach where you can combine general components to create each page.

msecret:
Just saw your comment about this being a tech demo. In that case, you can keep the CSS as is for now, I can help edit it after.

cmc333333:
Yep, none of the CSS should stay -- if we like this approach, we'll need a better targeting mechanism than just wrapping everything in a div. I don't want to go too far down that road without more feedback.

PR #122, opened 2014-12-08
Ehnc project filter with html

msecret:
Changed the way we structure the HTML in the template. Before we were
just dumping all the projects at the end. Now I'm adding the projects
under their place.

Reasons:
We were making the assumption that less html and moving things around
the DOM with JS would mean better performance. I realized this might
be an incorrect assumption, considering the amount of DOM manipulations
so I decided to setup this way because it's the easier solution to
start out with. I can branch and create code the other way later to
compare.
- This means there will be duplicate html in the template.
- This means the JS will not have to link the projects to their parent
  issue, country and volunteer, meaning less JS.

TODO:
- I haven't formatted country yet as @cm said to wait til we have to
  do a full database reset.


msecret:
The build was an issue due to an external dependency, so it should be fixed when it runs again.

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -7,67 +7,88 @@
...
-    <tr class="js-detailItem" name="{{ project.title }}">
+    <tr class="collapsibleItem" id="collapsible-volunteer-{{ loop.index }}">
       <td>
-        <img src="{{ volunteerpicture }}" alt="{{ volunteername }}" />
+        <table></tbody>
+          <tr name="{{ project.title }}"
+              aria-hidden="true">
+            {{ render_project(project, loop.index) }}
+          </tr>
+        </tbody></table>
```
cmc333333, line 78: `<table></tbody>` and `</tbody></table>` doesn't look right. Are sure you want to have a nested table?

cmc333333:
We'll sync up on country funds and issues -- you should be able to use both without confusion now.

I'm worried by the nested tables -- there seem to be many, and I can't figure out why they're needed.

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret, line 78: Wow good catch, it was supposed to be opening <tbody> tag.

msecret:
So for the tables:

Looking at line 34, I'm having this whole tr collapsed, and with the <tr> is both the campaign preview as well as each project belonging to the issue.

If I didn't create a new table and was at the <td> level here, how would I separate the campaign preview and the various projects belonging to the issue. I think each would have to be its own <tr>. I forgot a lot about writing html tables, so you may no better about nesting then me.

Looking at line 72, I found that I could remove that one nested table!

peacecorps/peacecorps/static/peacecorps/css/compiled/landing.css
```diff
@@ -0,0 +1,72 @@
+/*
+Error: Undefined mixin 'span-columns'.
```
cmc333333, line 2: I think this file can be safely deleted

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -7,67 +7,87 @@
...
-  {% endfor %}
-  {% for country in countries %}
-    <tr class="js-filteredItem"
-        data-filter-type="country"
-        name="{{ country.name }}">
+    {% if issue.featuredprojects.all().count() > 0 %}
+    <tr class="collapsibleItem" id="collapsible-issue-{{ loop.index }}">
       <td>
-        {{ country_map(country.code) }}
+        <table></tbody>
```
cmc333333, line 34: Another closing tag typo, I think

cmc333333:
If I understand this correctly, you have an outer table with rows for issues, volunteers, and countries. Each of those rows has an inner table, with the data relevant to that particular issue/volunteer/country.

The way I interpret the mockups, selecting an issue _replaces_ the issue selection table with a project selection table. I think you're seeing it as the issue's row "expands" (but is still surrounded by other issues). Is that right? @ericronne can your provide some clarity?

In any event, the "outer" table certainly isn't representing tabular data -- it's just a series of lists. The "inner" table comes closer, but we could argue that those aren't really tables either. Are we just using tables for making layouts?

msecret:
I see what you mean now. TI was thinking the rest of the issues were out of site in the mockup, but there is a <<issues link which means it should replace the content.

I also see what you mean about the table. I think we were talking about a table when we were discussing this so it stuck in my head, but the data doesn't lend itself that well to a table. It wouldn't be hard to change the whole thing to <divs> I might do that tomorrow.

ericronne:
Sorry I'm not an html guy so i need some handholding. Reach out to me when you get a sec and i'll help suss it out. If there's something in the design that i can change to make it more straightforward, i'll do that.

cmc333333:
@ericronne will do

@msecret Cool. Shall I merge this and we'll change the template in a later PR?

PR #123, opened 2014-12-08
Abstracts and return of static content

cmc333333:
- Abstract field was added to projects. It's just a text field at the moment.
- Templates make use of a method which returns either that field or a shortened first paragraph of the project's description
- Vignettes on landing page were removed and replaced with static content in the templates
- landing-page specific styles were removed

CSS classes were chosen with "meh" in mind -- just aiming to get the content present.


peacecorps/peacecorps/models.py
```diff
@@ -256,6 +258,23 @@ class Project(models.Model):
...
+        """If an explicit abstract is present, return it. Otherwise, return
+        the formatted first paragraph of the description"""
+        if self.abstract:
+            return self.abstract
+        elif self.description:
+            for block in json.loads(self.description)['data']:
+                if block.get('type') == 'text':
+                    data = block['data']
+                    # Naive string shortener
+                    if len(data['text']) > 256:
```
msecret: The only comment I'd have is to have 256 be a constant or a var so its easily editable. Not a big deal though.

peacecorps/peacecorps/models.py
```diff
@@ -256,6 +258,23 @@ class Project(models.Model):
...
+                        trimmed = data['text'][:256]
+                        trimmed = trimmed[:trimmed.rindex(' ')]
+                        data = {'text': trimmed + '...'}
```
msecret, line 275: I'm wondering if we should actually render the "..." in css. I was planning no doing this in other places. Can you think of reasons why doing it in CSS would worse?

peacecorps/peacecorps/templates/donations/donate_landing.jinja
```diff
@@ -75,8 +87,31 @@
...
+    <div class="u-hv_center--block">
+      <p>Brief blurb touting the search and sort tool, lorem ipsum dolor amet
+      lorem ipsum dolor.</p>
+    </div>
+  </section>
 </section>
 
+<section class="section">
+  <section class="section__content full_banner full_banner--neutral">
+    <h1 class="t-tittle--1">About Donating</h1>
```
msecret: t-title--1

msecret:
Whats the reason for the media def? I was curious, I don't see any problem with it.

peacecorps/peacecorps/models.py
cmc333333: Good call; will add it as a quick setting

peacecorps/peacecorps/models.py
cmc333333, line 275: We can do that, but it'll require we notify the template that the string has been shorted. Otherwise, there'll be an ellipses on all of the text, which would be misleading. LMK if you think that should be implemented or if this is simpler.

peacecorps/peacecorps/templates/donations/donate_landing.jinja
cmc333333: Good eye, thanks!

cmc333333:
On `media`: Django thinks about assets in two ways, `static`, meaning "with the code" and `media`, meaning user-uploadable. Generally, `media` assets keep track of their own locations/urls. In this case, the image isn't part of our static assets yet (if ever), so I'm referencing a file in `media`, by adding the system's default media location. We'll want to get a good understanding of how this content will be managed to know if this is an incredibly dumb idea :)

peacecorps/peacecorps/models.py
msecret, line 275: In that case, it sounds better to do it here.

PR #124, opened 2014-12-09
Moving Account FKs to point to code field

annalee:
The Peace Corps team uses account codes as their primary way of identifying projects. Pointing the Account foreignkeys on Campaign and Project to the 'code' field allows us to pull the code without having to do a table join.

This PR also includes a cleanup to get SirTrevor fields to serialize correctly.

### DIFF
--- a/peacecorps/peacecorps/fields.py
+++ b/peacecorps/peacecorps/fields.py
@@ -4,6 +4,9 @@ from django.conf import settings
 from django.db import models
 import gnupg
 
+from sirtrevor.fields import SirTrevorField
+import json
+
 
 class GPGField(models.Field, metaclass=models.SubfieldBase):
     def __init__(self, gpg_check=False, *args, **kwargs):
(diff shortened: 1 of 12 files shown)


peacecorps/peacecorps/fields.py
```diff
@@ -56,3 +59,12 @@ def get_prep_value(self, value):
...
         else:   # No GPG; just stick it in the DB
             return plain_text
+
+
+class BraveSirTrevorField(SirTrevorField):
+
+    def value_to_string(self, obj):
+        sirtrev = self._get_val_from_obj(obj)
+        sirjson = json.loads(sirtrev)
+        sirjohn = json.dumps(sirjson)
```
cmc333333: Would it work if this was replaced with `return str(sirtrev)`? I don't think we need to parse and encode the data.

cmc333333:
This includes some conflicts with migration numbers; `master` is up to 37, but this starts at `36`. Would you mind combining those migrations into a single migration 38? You just need to concatenate all of the `operations` fields.

annalee:
@cmc333333 Django Migrations doesn't actually care about the file names--those are just for our reference. Once of the included migrations is a merge between these migrations and master's, so there shouldn't be a conflict. Is there a particular reason we want to maintain strict sequential numbering?

cmc333333:
My bad, I didn't realize it was so smart!

cmc333333:
Thanks @annalee !

annalee:
You're correct about str() though. I thought I'd tested that and it was escaping things weirdly, but apparently not.

PR #125, opened 2014-12-09
Ehnc collapsible js

msecret:
Builds on #122 

JS for collapsible.

Collapsible will hide something and give it controls to open it up
again.

This is an iteration of this design. A further improvement could be to
include this code in the discover app, to limit global access to the
DOM.


cmc333333:
JS looks okay -- thanks for the tests. Once you merge in upstream, landing.(s)css should be gone.

msecret:
Replaced with #129

PR #126, opened 2014-12-09
Ehnc finish filter

msecret:
Builds on #125

Finished up filter code

Finished it up by adding the click events for the nav.

TODO:
- Possibly rename at some point.


msecret:
Replacing with #129

PR #129, opened 2014-12-09
Ehnc filter new direction

msecret:
Changed html structure and js code for project sorter

Changed htlm structure to not be table as discussed.

This is not completely to design specs yet. Currently, its missing the
first "page" on the murally. This means that the campaign fund and all
associated projects are there and are filterable with the nav, but not
the single line titles of the countries and issues.

I also haven't done a separate rendering for the volunteer vs country
vs issues project yet, meaning they all have the same information and
layout right now.


peacecorps/peacecorps/static/peacecorps/css/compiled/landing.css
```diff
@@ -1,72 +1,54 @@
+/* Neat 1.7.0
```
cmc333333: I think this file can be completely removed, yes?

peacecorps/peacecorps/views.py
```diff
@@ -229,9 +229,9 @@ def donate_projects_funds(request):
     """
     The page that displays a sorter for all projects, issues, volunteers.
     """
-    countries = Country.objects.all()
-    issues = Campaign.objects.filter(
-        campaigntype=Campaign.SECTOR).order_by('name')
+    countries = Campaign.objects.filter(
+        campaigntype=Campaign.COUNTRY).order_by('name')
+    issues = Issue.objects.all()
```
cmc333333: May want to order these by `name` or even `pk`, just to provide consistency.

:w
```diff
@@ -0,0 +1,288 @@
+from urllib.parse import urlencode
```
cmc333333: I don't think this file was supposed to be committed. Yay Vim, though

peacecorps/peacecorps/static/peacecorps/css/src/landing.scss
```diff
@@ -0,0 +1,22 @@
+@import "bourbon/bourbon";
```
cmc333333: This file should be removed, too

PR #130, opened 2014-12-10
New design of landing page

msecret:
Added grid component for basic grid work. Starting to get rid of
css showcase class. Fixed neat by moving grid-settings into own file
and including it before neat.

TODO:
- Make grid component filename underscored.

PR #131, opened 2014-12-10
Start on new designs of project page

msecret:
I have grid working in a different file, so left that out for now.


cmc333333:
My mistake; I should have merged #130 first.

PR #134, opened 2014-12-11
Ehnc project page design

msecret:
This finishes up the project page design.
- I changed the python form to support the different form of options. I didn't check this very much yet.
- Set some max image widths as discussed so images will never be larger then a certain width.
- Added some inline grids
- Sir trevor styling. I think this can be done in the sir trevor code, but I found it faster to do it like this for now. Maybe I should add a todo if you think it'd be better to do this in sir trevor code.

Screenshots
screen shot 2014-12-10 at 5 34 12 pm
screen shot 2014-12-10 at 5 34 25 pm
screen shot 2014-12-10 at 5 34 39 pm
screen shot 2014-12-10 at 5 35 11 pm
screen shot 2014-12-10 at 5 35 40 pm
screen shot 2014-12-10 at 5 35 59 pm
screen shot 2014-12-10 at 5 36 17 pm
screen shot 2014-12-10 at 5 36 26 pm


cmc333333:
This conflicts with the merged #133, otherwise I think it's good. Globally styling all sir trevor blocks will work for now, but I've been thinking about a longer term strategy. I think we could extend it pretty easily so that `project.content.html("project-landing")` would look for overridden templates in the "project-landing" directory for each of its blocks.

peacecorps/peacecorps/fixtures/tests.yaml
```diff
@@ -1,5 +1,5 @@
 - fields: {category: mem, code: DCC-0123, community_contribution: null, current: 0,
-    goal: null, name: Stephanie Brown}
+    goal: 63000, name: Stephanie Brown}
```
cmc333333: This isn't right. Memorial funds, Country funds, etc. don't have a goal

peacecorps/peacecorps/static/peacecorps/css/compiled/donation.css
```diff
@@ -1,78 +1,78 @@
...
-  -webkit-box-sizing: border-box;
-  -moz-box-sizing: border-box;
-  box-sizing: border-box; }
-
-*, *:before, *:after {
-  -webkit-box-sizing: inherit;
-  -moz-box-sizing: inherit;
-  box-sizing: inherit; }
-
+Error: Invalid CSS after "...ng project page": expected "{", was "}"
```
cmc333333: SCSS error

peacecorps/peacecorps/templates/donations/donate_project.jinja
```diff
@@ -51,17 +56,70 @@
...
           <strong>
             {{humanize_amount(project.account.current)}}
-          </strong>
+          </strong><br>
+          total raised
         </span>
       </div>
     </div>
 
+<<<<<<< HEAD
```
cmc333333: Merge issues

cmc333333:
Thanks and sorry for the merge conflicts

PR #139, opened 2014-12-11
Project sorter, nav

msecret:
Styles the nav as a blue banner with inline links. I decided to hide
the "Select a category..." text from the middle and small breakpoints
as it took up too much space.

For behavior, it will change the aria-selected role when a nav item
is clicked and use this for styling it as underlined.

Removed option in discover to specify which is selected as it has no
use at this point.

For screenshots: I don't have Archer installed on this computer yet, please ignore fonts.
screenshot from 2014-12-11 14 23 55


cmc333333:
Noice!

PR #142, opened 2014-12-12
Image Processing

annalee:
Function to save resized copies of images to serve to different sizes of device. This is just the saving side; the logic to choose which size to display will be in the PR with the 508-compliant sir trevor widget.


cmc333333:
This'll need to be integrated with S3, but it's a good first step.

PR #144, opened 2014-12-15
Resolves several country issues

cmc333333:
- Adds entries for Kosovo and "Eastern Caribbean"
- Fixed country code for Gambia
- Fixed Kyrgyz Republic name
- Add additional zoom levels (so that small countries don't have opaque boundaries)
- Account for translations when cropping paths out
- Add a script to simply zoom the map (used to generate the Eastern Caribbean image)

Github gave up on the diff due to all the SVGs, but there's a [gist](https://gist.github.com/cmc333333/e1c105da44dd03ea1591) of the relevant part


msecret:
Looks good

PR #149, opened 2014-12-16
Removing featured_image from the campaign model, since it's no longer used.

annalee:


peacecorps/contenteditor/admin.py
```diff
@@ -132,6 +129,7 @@ class FAQAdmin(SortableAdminMixin, admin.ModelAdmin):
 
 admin.site.register(models.Account, AccountAdmin)
 admin.site.register(models.Campaign, CampaignAdmin)
+admin.site.register(models.FeaturedCampaign)
```
cmc333333: Featured Campaign is supposed to be gone, hey?

annalee: yup. sorry. put it back in to fix the test fixture. pulling it back out.

PR #150, opened 2014-12-16
Resize to the cloud

cmc333333:
Moves resizing code into `contenteditor` app. Makes it work with the default storage (allowing it to send to S3)

These resized images aren't used anywhere yet, and I've opened a separate issue for making sure the correct libraries are installed.


msecret:
Sorry I didn't ask this before, but if the image sizes change due to design changes, would it be possible to resize existing images easily?

cmc333333:
We'd need to write a script to do this, but it shouldn't be too hard.

PR #153, opened 2014-12-16
508 version of the image widget for Sir Trevor

annalee:
Paired on this with @cmc333333. It takes a title and description as well as an image, and uses them to populate the title and alt text for the image when displaying the sirtrevor block on the site.


peacecorps/contenteditor/static/contenteditor/js/image508block.js
```diff
@@ -0,0 +1,65 @@
+SirTrevor.Blocks.Image508 = SirTrevor.Block.extend({
+  type: 'image508',
+  title: 'Image508',
```
cmc333333: I remember you wanted to change this to "Image". Does that break anything?

peacecorps/contenteditor/static/contenteditor/js/image508block.js
```diff
@@ -0,0 +1,65 @@
...
+  droppable: true,
+  uploadable: true,
+  formattable: false,
+
+  icon_name: 'image',
+
+  editorHTML: function() {
+    return [
+      '<br  />',
+      '<label class="st-input-label">Image Title</label>',
```
msecret, line 13: If we're focusing on accessibility here, is there a reason we would put the label "for" attribute?

seanherron, line 13: Yeah, these should have `for` attributes, but I don't think that is blocking for a merge.

cmc333333:
We can add "for" labels shortly; merging for lack of time

PR #155, opened 2014-12-17
Language Additions

seanherron:
Basically, this goes through and removes Lorem Ipsum in favor of slightly bland but ready for revision text.


cmc333333:
Well that's less fun.

PR #158, opened 2014-12-17
Account for boto bug

cmc333333:
This hacks around https://github.com/boto/boto/pull/2637 so that our demo site's static assets can appear on the admin site.


seanherron:
LGTM, let's try this!

PR #159, opened 2014-12-18
ehnc First pass at improved project sorter

msecret:
Quick pass at implementing the project sorter. Uses some new methods
on discover to hide elements, show them, whatever.

It's performing very very slowly right now, essentially takes a few
seconds for the elements to even hide. Something we may want to be
concerned about.

TODO:
- Use aria attributes to hide
- Unit tests
- Styling
- Performance


peacecorps/peacecorps/static/peacecorps/js/src/discover.js
```diff
@@ -24,6 +26,10 @@ Discover.prototype.init = function(root, $navLinks) {
       $selectedLink;
 
   this.$navLinks = $navLinks;
+  this.$pages = this.$(this.ccPage);
+  console.log('pages: ' + this.$pages.length);
+  this.$pageLinks = this.$(this.ccPageNavs);
+  console.log('page links: ' + this.$pageLinks.length);
```
cmc333333, line 32: Probably don't want these debug statements

cmc333333:
I wonder why it's so slow -- nothing stands out to me.

Waiting until after the demo tomorrow, then will merge.

PR #161, opened 2014-12-18
Issue icons

cmc333333:
- Update issues fixture to reference new icon set
- Grey icons for featured projects on the home page via image-mask
- Macro to use a field's "url" property if present and an empty string otherwise

Both S3 and the shared media repository have been updated.


peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -107,7 +108,7 @@
     <img class="u-pullr"
       class="section__content--no_space_r"
       style="width: 42%;"
-      src="{{ media('World-by-Ana-Maria-Lora-Macias-CC-BY.png') }}"
+      src="{{ url_or(issue.icon, "") }}"
```
cmc333333, line 141: I don't think this would ever be null, so we may remove this macro at some point.

peacecorps/peacecorps/templates/donations/includes/macros.jinja
```diff
@@ -0,0 +1,7 @@
+{% macro url_or(field, default) -%}
```
msecret, line 1: I really like this inclusion of a macro file.

peacecorps/peacecorps/static/peacecorps/css/src/components/_bg_img.scss
```diff
@@ -20,13 +20,14 @@
...
-    margin-left: 1.5em;
-  }
+  display: none;
+}
+.bg_img--icon:not(old) {
+  display: inline-block;
+  width: 1.5em;
+  height: 1.5em;
+  background-color: #888;
+  -webkit-mask-size: 1.5em 1.5em;
```
msecret, line 32: The use of the "-webkit" prefix here without a moz prefix or just a generic one seems to point to this property not being available in non-webkit browsers. Are we just accepting that for this iteration?

cmc333333, line 32: Some context: this uses an image (the icon svg) as a mask, to cut out a grey image in the shape of the icon.

Many browsers (not IE) support `mask`ing, but only webkit supports a mask with a size.

Do you think we should try embedding the svg and using a custom style sheet?

PR #162, opened 2014-12-18
ehnc Styling for sorter

msecret:
The new header sections have been styled according to mockups.

Spacing could still use some work, I might leave that for another
iteration as I believe this is usable.

The code is getting messier and messier, I think it needs a refactor
soon which I've already been thinking about.

screen shot 2014-12-18 at 3 28 59 pm
screen shot 2014-12-18 at 3 29 05 pm


cmc333333:
In terms of markup, it might be helpful to start `include`ing sub-templates rather than having so many macros.

PR #169, opened 2014-12-22
Global/General fund fixtures & css tweaks

cmc333333:
- Add fixtures for global/general fund images
- Switch to using funds' _icon_ instead of _featured_image_
- Add a CSS class for masks which will expand the mask to fill the width of its container while keeping square

Images have already been uploaded to s3 and the media directory. Looks like:

screen shot 2014-12-22 at 12 13 38 pm
screen shot 2014-12-22 at 12 13 45 pm


peacecorps/peacecorps/static/peacecorps/css/src/components/_mask.scss
```diff
@@ -16,6 +16,13 @@
   height: 10em;
 }
 
+/* Make a square with height/width equal to container's width */
+.mask--fill-square {
```
msecret: mask--fill_square to keep in line with somewhat bad naming convention

PR #171, opened 2014-12-23
WIP ehnc sytling donation

msecret:
Don't merge just review.

The new design for the form changes some of the radio multi select
fields into checkboxes, we should decide if we want to change this in
the python code right now.

TODO:
- fix python unit tests.
- style dedication form.
- behavior with js.


peacecorps/peacecorps/templates/donations/donation_payment.jinja
```diff
@@ -6,138 +6,168 @@
...
+            <label for="id_payer_name" class="form--required form__field__label">
+              full name</label>
+            <div class="form__field__input">{{form.payer_name}}</div>
+        </div>
+
+        <div class="form__field">
+            {{form.billing_address.errors}}
+            <label for="id_billing_address" class="form--required form__field__label">
+              street address</label>
+            <div class="form__field__input">{{form.billing_address_extra}}</div>
```
cmc333333: Should this be `form.billing_address`?

peacecorps/peacecorps/templates/donations/donation_payment.jinja
```diff
@@ -6,138 +6,168 @@
...
         </div>
 
-        <div>
+        <div class="form__field">
             {{form.billing_address.errors}}
-            <label for="id_billing_address" class="required">Street Address</label>
-            <div class="value">{{form.billing_address}}</div>
+            <label for="id_billing_address"
+                class="form--required form__field__label">
```
cmc333333: I don't think this one is required, and should it be `id_billing_address_extra`? Since the backend is not hooked up yet, would you mind commenting this out for now?

peacecorps/peacecorps/static/peacecorps/css/src/components/_form.scss
```diff
@@ -9,3 +17,74 @@
     display: none;
   }
 }
+
+form ul {
+  @extend .u-unmarked_list;
+  @extend .u-inline_list;
+}
+
+.form--required {
```
cmc333333, line 26: If this class has changed, you'll need to update the JS accordingly. It sets certain labels as "required" when others are selected/unselected.

peacecorps/peacecorps/templates/donations/donation_payment.jinja
```diff
@@ -6,138 +6,168 @@
...
 
-        <div>
+      <fieldset>
+        <legend><h2>Payment Method</h2></legend>
+        <div class="form__field">
             {{form.payment_type.errors}}
-            <label for='id_payment_type' class="required">Payment Type</label>
-            <div class="value">{{form.payment_type}}</div>
+            <label for='id_payment_type' class="form--required form__field__label">
+              </label>
```
cmc333333, line 114: Looks like this label isn't needed anymore?

peacecorps/peacecorps/forms.py
```diff
@@ -41,6 +41,8 @@ class DonationPaymentForm(forms.Form):
     country = forms.ModelChoiceField(
         queryset=Country.objects, to_field_name='code', initial='USA')
     billing_address = forms.CharField(label="Street Address", max_length=80)
+    billing_address_extra = forms.CharField(label="Street Address (cont)",
+        max_length=80)
```
cmc333333: If this is optional, you'll want to add `required=False`. I'd bet that's what's triggering the unit test failures

cmc333333:
Changing the radios into checkboxes will require a few modifications to tangled code, but shouldn't be too bad. I'll dive a little deeper in the morning.

cmc333333:
An additional todo: add the new address field to the pay.gov xml. This is a good first step, though.

PR #173, opened 2014-12-23
ehnc Dedication subform in donation form

msecret:
Styling was put together from existing form style. Haven't looked at it
extensively.

TODO:
- Ensure design element of separting subform works.
- We probably want to change the information consent radio field to
  a checkbox like the rest.

screen shot 2014-12-23 at 12 12 38 pm


cmc333333:
Looks good!

PR #175, opened 2014-12-24
ehnc Donation form errors

msecret:
Styled errors and some more layout things. Basic responsive
working.

Decided not do do the bubble UI for error messages for this
iteration due to time constraints.

For further dev; I think Eric wanted form errors to appear
as you type in the form, which also may be too much for
this iteration.

TODO:
- js
- Bug on "business with peace corps" checkbox where label
  is not linked to checkbox (even though I did correct for
  attr), will figure out later.
- address cont field is showing error when main address
  field has error.

screen shot 2014-12-23 at 5 21 08 pm


peacecorps/peacecorps/forms.py
```diff
@@ -10,6 +10,11 @@
 class DonationPaymentForm(forms.Form):
     """Collect contact information and dedication information about a donor"""
 
+    default_errors = {
+        'required': 'This field is required',
+        'invalid': 'Enter a valid value'
+    }
```
cmc333333: I don't _think_ this does anything?

msecret: Correct forgot to take it out.

PR #176, opened 2014-12-24
Account for projects with no volunteer photo

cmc333333:
- Add a new macro for volunteer photos
- This macro shows the image url if present, and a div with the name/state if not
- Make the text smaller when in the volunteer selection screen
- Typo in peace corps url

screen shot 2014-12-24 at 10 40 21 am
screen shot 2014-12-24 at 12 43 36 am
screen shot 2014-12-24 at 12 43 57 am


ericronne:
Looks like i may have misread here yesterday and consequently mislead you. In the collapsed volunteer sorter listing, we should show the map when there is no volunteer photo, like so ...
image

see mural. thx

cmc333333:
Updated to match the design -- map of the country if in the collapsed sorter.

PR #177, opened 2014-12-25
JS donation form

msecret:


peacecorps/peacecorps/static/peacecorps/js/src/donation.js
```diff
@@ -67,12 +68,13 @@ var Init = {
...
 $().ready(function() {
-  var $discoverApp = $('.js-discoverApp');
+  var $discoverApp = $('.js-discoverApp'),
+      $form = $('.js-form');
+
   // TODO I want to rebuild the Init class to remove this check at some point.
   if ($('.landing').length < 1) {
     Init.donorTypeFields();
     Init.countryRequirements();
-    Init.dedicationFields();
```
cmc333333, line 75: I think you can get rid of the whole function, too?

PR #181, opened 2014-12-29
Migrationcleanup

annalee:
Updates tests/test data to bring them in line with current migrations, and setting the remove_account_id migration to handle null values in project overflow fields.


cmc333333:
I think it's slightly better style to use a guard that explicitly checks the case you care about (here, the field being None) rather than a side effect (here, the exceptions). That said, this is just a one-off, so I'm okay merging.

PR #188, opened 2014-12-30
Donation form top part

msecret:
Renders a version of the top of the donation form for both projects
and funds.

I was messing around with the best way to pass the project id to
the form to get the current project. CM and I discussed having a
different url and using the slug, but in the end I ended up with this
because its easiest and it worked. I'd be happy to look at different
ways of doing it though, I don't really like that different outputs
happen with the inclusion of a query param.

@ericronne , its difficult at this moment to get the "your donating"
part onto the same line as the 3 circles because the circles are shared
code with the rest of the site, so making that work would be a larger
architectural change. Also: the layout is off, the left margin doesn't
match whats below. We should talk about this because its due to grid
inconsistencies. It will be easy to fix once we have a grid.

screen shot 2014-12-29 at 4 55 03 pm
screen shot 2014-12-29 at 4 55 17 pm


peacecorps/peacecorps/views.py
```diff
@@ -32,6 +33,12 @@ def donation_payment(request):
     if not account:
         return HttpResponseBadRequest('Invalid project')
 
+    project = None
+    if project_id is not None:
+        project = get_object_or_404(
+            Project.published_objects.select_related('country', 'account'
+            'volunteerpicture', 'featured_image'), id=project_id)
+
```
cmc333333, line 37: You shouldn't need to pass around the project id. Try this instead:

``` python
project = account.project_set.first()
```

msecret, line 37: If there are multiple projects, is there a case where there would be more then one?

cmc333333, line 37: Shouldn't be, the `account` field is unique on `Project`. I don't _think_ `project_set` would include `Project.overflow`, since it has a `related_name`, but I haven't tested.

PR #190, opened 2014-12-30
Link to specific FAQ from donation form (checks)

cmc333333:
- Adds slug to FAQ model
- Uses this slug when generating anchor links on the FAQ page
- Updated fixtures to include latest FAQ content (and slugs)
- Non-JS users get all FAQs open by default
- If a url hash is present, it is checked when initializing the collapsibles. If there's a match, that collapsible stays open
- Link to the specific question from the donation form


peacecorps/peacecorps/templates/donations/faq.jinja
```diff
@@ -15,14 +15,14 @@
 <section>
   <ul>
     {% for faq in object_list %}
+      {% set anchor = faq.slug or "faq-" + faq.pk|string %}
       <li>
         <header>
           <a aria-expanded="false" href="#"
-             aria-controls="collapsible-{{faq.pk}}">{{faq.question}}</a>
+             aria-controls="{{anchor}}">{{faq.question}}</a>
```
msecret, line 22: I will check with Nick to see if the "collapsible" part is important for aria. It was actually something I found when learning about aria, not something I made up myself.

PR #192, opened 2014-12-31
Higher-contrast colors for the django admin

annalee:
This is a first pass at higher-contrast colors for the admin. The original admin interface has a few gradients for the header bars etc that aren't duplicated here because I don't have the software, but I think we can merge without them and add them later.


cmc333333:
@annalee - thinking about this a little more, if we're replacing whole CSS files, why do we also need to alter the templates?

PR #182, opened 2014-12-29
Resizing images on the media model (Issue #179)

annalee:
Resolves https://github.com/18F/peace-corps-product-management/issues/179

As with the SirTrevor image processor, creates and saves resized copies of any images uploaded to the Media model. Currently saving to the same place Sir Trevor saves--I'll be removing this functionality from Sir Trevor's processor once its images are being properly saved to the media model.

Question: This is currently pulling filenames for the resized images from the title field. Any preference to generate filenames a different way?

PR #193, opened 2014-12-31
Savesirtrevor

annalee:
Resolves https://github.com/18F/peace-corps-product-management/issues/177 . Saves images from Sir Trevor to the Media model.

(Needs a test before merge).


annalee:
I've merged https://github.com/18F/peacecorps-site/pull/182 into this PR, since they're dependent on each other.

#182 adds image processing to the Media model and this PR removes the Sir Trevor image processor so that images saved through sir trevor won't get processed twice.

peacecorps/peacecorps/models.py
```diff
@@ -7,12 +9,39 @@
...
 from peacecorps.fields import GPGField, BraveSirTrevorField
 
 
+def imagesave(description):
+    """Saves images from Sir Trevor fields to the media model."""
+    if not description:
+        # if description is empty for any reason, it has no images.
+        return False
+
+    description = eval(description)
```
cmc333333: `eval` is almost always a bad idea. Did you mean to parse the json instead?

peacecorps/peacecorps/models.py
```diff
@@ -7,12 +9,39 @@
...
+            imagepath = block['data']['file']['path']
+
+            desc = block['data']['image_description']
+
+            thisimage = Media(
+                title=title,
+                file=imagepath,
+                mediatype=Media.IMAGE,
+                description=desc,)
+            thisimage.save()
```
cmc333333, line 39: Won't this create new `Media` models every time the project/campaign is saved -- even if the images are identical?

cmc333333:
Can we take a quick step back to document why adding a `Media` instance for each embedded image is helpful? I imagine it'd a) centralize image resizing code and b) allow all images to be queried at once (e.g. for an image library). Is there more?

Unrelated - we've already got `pc_logo.png` in the repo. Can it just be copied for that test?

annalee:
@cmc333333: the media library is pretty much the main reason, though centralizing the resizing code is nice, too. Do you want me to add a comment to that effect to the function?

As for the images--I'm trying to get at that image, but it's taking the default_storage path from the model (which is setting it to RESIZED_IMAGE_UPLOAD_PATH), and Django is throwing a SuspiciousFileOperation when I try to go up the directory tree to get to that image.

PR #201, opened 2015-01-03
Sorter, project, fancy icon

msecret:
Change icon so the background is in circular mask.

I've been struggling to make the circle mask based off percentages
rather then setting the width and height in ems. I tried some solutions
but nothing worked well so will work on that part after.

https://github.com/18F/peace-corps-product-management/issues/307


ericronne:
The circle is great, thx. It should be outlined in AAA698, btw, not the current gray color (888888).

PR #202, opened 2015-01-05
Color Issue Icons

cmc333333:
Modifies SVG icons to be filled with predefined colors. As these icons are editable, this has to be done dynamically.

Specifics:
- Restricts icon on Issue model to be an SVG (with a width and height)
- Adds configuration for colors. All uploaded icons will be converted to a square viewport and all stroke/fill colors will be replaced with these configured colors
- Adds an access method to find a specific color for the icon (e.g., "give me this icon in grey")
- Adds a management command to re-build all colored icons (useful should the colors change)
- Uses the grey icon on the landing page (featured projects) and in issue selection on the sorter
- Uses the white icon on the sorter, once an issue has been selected

Updated icons are already on S3 and in the media repo. ~~Only weird one is the soccer ball, which has a strange transparency issue.~~

screen shot 2015-01-05 at 12 24 52 pm
screen shot 2015-01-05 at 12 25 22 pm
screen shot 2015-01-05 at 1 19 08 pm
screen shot 2015-01-05 at 12 25 40 pm


ericronne:
I'm fixing the soccer ball .svg now...

ericronne:
Try this soccer ball for youth...
https://drive.google.com/a/gsa.gov/file/d/0BwWYNZcEDfwaQnZjcVhjUnN1T28/view?usp=sharing

cmc333333:
Thanks @ericronne - I've updated the screenshots and S3/the media repo.

peacecorps/peacecorps/templates/donations/donate_landing.jinja
```diff
@@ -51,13 +51,15 @@
...
         <div class="grid--40 u-clearfix">
-          {% if project.issue() %}
+          {% set issue = project.issue() %}
+          {% if issue %}
             <div class="issue">
               <img class="bg_img--icon"
-                   style="-webkit-mask-image: url('{{ url_or(project.issue().icon, "") }}');"
-                   alt="{{project.issue().name}}"
+                   alt="{{issue.name}}"
+                   style="height: 1.5em; width: 1.5em;"
```
msecret: Do we have to add absolute widths and heights here? It will break responsive layouts.

peacecorps/peacecorps/issue_icons.py
```diff
@@ -0,0 +1,133 @@
+from copy import deepcopy
```
msecret: I'm not sure this is the best location for this file. What about a util folder? Could this be considered svg utility?

peacecorps/peacecorps/issue_icons.py
```diff
@@ -0,0 +1,133 @@
...
+    while match:
+        if match.group('value').strip() != 'none':
+            new_style = haystack[:match.start()] + 'fill: ' + hex_val
+            new_style += haystack[match.end():]
+            haystack = new_style
+        match = FILL_STYLE.search(haystack, match.start() + 1)
+    return haystack
+
+
+def color_icon(svg_xml):
```
msecret: How long is this code taking to run for an icon?

peacecorps/peacecorps/templates/donations/donate_landing.jinja
cmc333333: I'll play with it a bit. We can probably do a background image instead.

peacecorps/peacecorps/issue_icons.py
cmc333333: Can do.

peacecorps/peacecorps/issue_icons.py
cmc333333: Negligible on the files I've tested. I've found some bugs testing the map of the globe, but I'll try it there to provide a more useful metric.

cmc333333:
Updated by renaming/moving svg utility file, making the issue icon a bg image which scales with the text, and being more accepting of SVG file formats.

Tested locally using the globe map (~1.5M) and it runs instantaneously (though there may be a delay when uploading to S3).

PR #204, opened 2015-01-06
Header markup updates + redirect

cmc333333:
There are two components to this PR, adding a redirect & header for payment failures and markup/style changes to make that happen. Once I started tweaking the base header section, I noticed some issues with header image size. This PR moves the "beta" warning, the Peace Corps logo + title text, and image into a single container and strips out some absolute positioning. Custom images now open to 85% the height of the screen while maintaining aspect ratio.

It also displays a warning when the user's payment fails. Fonts are wrong, but that's a site-wide issue.

screen shot 2015-01-06 at 10 09 06 am
screen shot 2015-01-06 at 10 09 17 am
screen shot 2015-01-06 at 10 10 24 am
screen shot 2015-01-06 at 10 11 16 am
screen shot 2015-01-06 at 10 11 41 am

Failure screen:
screen shot 2015-01-06 at 10 18 44 am


peacecorps/peacecorps/static/peacecorps/css/compiled/donation.css
```diff
@@ -2262,6 +2226,36 @@ html {
...
+ * Z-Index
+ */
+.top_image {
+  border-bottom: 0.75em solid #F2EFE8;
+  background-repeat: no-repeat;
+  background-position: center bottom;
+  margin-bottom: 2em; }
+
+.top_image--default {
+  background-image: url(../../img/default_top.png); }
```
msecret, line 2251: Should this not be /static?

msecret, line 2251: wait I realize now: we can't do that because it we don't have the static() function from python here. nvm.

msecret:
Cool I loved the way you did this!

PR #205, opened 2015-01-06
JS users immediately submit review form

cmc333333:
Ideally, JS users will never see the payment confirmation screen. As a first step, this will submit the confirmation form for the user if JS is enabled. This PR also modifies the submit button text if JS is enabled.


peacecorps/peacecorps/static/peacecorps/js/src/form.js
```diff
@@ -56,6 +56,8 @@ var initForm = function($form) {
         switchGroupIds[i],
         $form);
   }
+
+  $form.find('button[type="submit"]').text('Continue to Pay.gov');
```
msecret, line 60: Ideally this should use a "js" prefixed css class, but I don't really care in this context.

PR #206, opened 2015-01-06
Add icon font, font-awesome

msecret:
Setup:
- includes font-awesome as an npm module.
- cps the sass files from font awesome into a convenient lib folder
  under css src.
- copies the font-awesome fonts into a convenient fonts folder in root.
- modifies the font-awesome sass vars to point to the correct /fonts
  folder. (this is the part I dislike the most, but sass will not compiled
  correctly if I put the var in my code).
- load font-awesome css async through ajax

note: there's two ways to include font-awesome in this way: including
it in donation.css, the main css file, or including it as a separate
file, as I've done now. I picked my way because fonts probably aren't
required at start so shouldn't block critical rendering path. BUT,
further down the road, when we have a build task to remove all unused
css classes from the output, it may be better to move this to donation
css as it will be less file size in total. the biggest problem with
current way is stupid file-path var fix that I'm using in grunt.

Another thing to note: if we weren't checking in front-end dependencies
then this would not add as much weight to the repo.

@meiqimichelle @shawnbot @noahmanger , you guys can review and comment.


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -154,6 +155,17 @@
...
+<script type="text/javascript">
+(function() {
+  var css = document.createElement('link');
+  css.href =
+    '{{ static("peacecorps/css/compiled/lib/font-awesome/font-awesome.css") }}';
+  css.rel = 'stylesheet';
+  css.type = 'text/css';
+  document.getElementsByTagName('head')[0].appendChild(css);
+})();
+</script>
```
cmc333333, line 174: What do you think about adding this to the header inside a `<noscript>` tag as well?

cmc333333:
We've already got precedent for SASS libraries not being part of the code base (bourbon, neat) -- could we do the same here? As long as the compiled CSS is checked in, non-node people can still work on the code base.

msecret:
We could easily have css/src/lib be gitignored, so it wouldn't be checked in. But then sass compilation would fail if you don't npm install.

I'm might be moving in a different direction where I want to completely remove ruby as a dependency, replacing with node/npm, and then not check in libs. I want to have a discussion if you don't agree with that.

cmc333333:
I agree, having some requirements in npm and others in ruby doesn't make sense. Removing libs from the code base is also good! The compiled CSS is a bit different (it reduces dependencies even further), but your approach sounds like a good strategy.

peacecorps/peacecorps/templates/donations/base.jinja
msecret, line 174: Good idea, done.

PR #208, opened 2015-01-07
ehnc move css compilation to libsass

msecret:
This will remove the ruby dependency.

The only problem currently is omega isn't setting a margin-right of
0. This isn't currently causing problems with the layout. I may
come up with my own work around for this if its still a problem.


cmc333333:
Excellent. Would you mind updating the README to use `grunt sass`?

cmc333333:
Thanks!

PR #209, opened 2015-01-07
Funded/Not Funded column for PCPPs in the admin.

annalee:
Resolves https://github.com/18F/peace-corps-product-management/issues/265
- [x] On the Project model Admin List view, introduce an additional column with "Funded Status" (either Funded or Not Funded)
- [x] On the Project model Admin Detail view, in the `Account Info` panel, include "Funded Status: (Funded/Not Funded)" in text (not editable)
- [x] On the Featured project front pages model Admin List view, introduce an additional column with "Funded Status" (either Funded or Not Funded

### DIFF
--- a/peacecorps/contenteditor/admin.py
+++ b/peacecorps/contenteditor/admin.py
@@ -55,7 +55,14 @@ class CampaignAdmin(admin.ModelAdmin):
 
 
 class FeaturedProjectFrontPageAdmin(admin.ModelAdmin):
+    list_display = ["project", "funded_status"]
     raw_id_fields = ['project']
+
+    def funded_status(self, obj):
+        if obj.project.account.funded():
+            return "Funded"
+        else:
+            return "Not Funded"
 
 
 class IssueAdmin(admin.ModelAdmin):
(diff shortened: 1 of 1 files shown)


peacecorps/contenteditor/admin.py
```diff
@@ -88,10 +96,17 @@ class ProjectAdmin(admin.ModelAdmin):
     raw_id_fields = ['account', 'overflow',
                      'volunteerpicture', 'featured_image']
     exclude = ['media']
+    readonly_fields = ['funded_status']
```
cmc333333, line 99: Does `FeaturedProjectFrontPageAdmin` need this line too? If not, does it need to be here, either?

annalee, line 99: The readonly_field only applies to the detail view. Per @seanherron's spec, we're only displaying funded status on the `FeaturedProjectFrontPageAdmin` list view.

cmc333333, line 99: Ah, thanks for the explanation

PR #212, opened 2015-01-08
Better error messages on donation amount

cmc333333:
New error messages for missing amount, < $1 amount, >= $10000 amount, and > remaining amount.

screen shot 2015-01-08 at 12 46 36 pm
screen shot 2015-01-08 at 12 46 48 pm
screen shot 2015-01-08 at 12 47 04 pm
screen shot 2015-01-08 at 12 47 18 pm
screen shot 2015-01-08 at 12 47 35 pm

### DIFF
--- a/peacecorps/peacecorps/forms.py
+++ b/peacecorps/peacecorps/forms.py
@@ -5,6 +5,7 @@ from django.core.exceptions import ValidationError
 from localflavor.us.us_states import STATE_CHOICES
 
 from .models import Country
+from .templatetags.humanize_cents import humanize_cents
 
 
 class DonationPaymentForm(forms.Form):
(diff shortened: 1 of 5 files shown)


msecret:
This all looks good. One thing I've been thinking for a while on these forms is what if we have a design requirement to have certain or all error messages on the form appear while typing? I could easily see Eric and the team wanting that, especially for the amount errors. Maybe we could brainstorm or discuss ways we would do that in the future.

cmc333333:
@msecret +1, it won't be too hard to have an api to validate the forms and return the same info

msecret:
The only other thing I thought of: we have an "error" red color that we're using on the main form. I was thinking we may want to use that color for the text here, although I'd want that decision to be made by @ericronne . I do see that you stuck to the mockups, just a design thing I see. I'll merge this anyway for now.

PR #223, opened 2015-01-12
Donation Amount Form JS + CSRF fix

cmc333333:
- No more radio select on donation form - the presets are direct links and the custom amount opens a larger text box
- No longer POST on the project page -- use GETs to the donation form instead
- For non-JS users, there are no presets, only the custom amount field
- For JS users, if there is no error, the form starts with the presets visible
- For JS users, if there was an error (or similar), the form starts with the custom amount visible
- Extended the existing form JS to account for shown/hidden elements which are triggered by anchors
- Added a few lines to specify fields which should be shown/hidden if JS is present

~~I'd like to add some JS tests, but could use feedback before.~~ Done

Non-JS:
screen shot 2015-01-12 at 12 41 35 pm

JS, no error:
screen shot 2015-01-12 at 12 41 52 pm
screen shot 2015-01-12 at 12 41 58 pm

JS, with error:
screen shot 2015-01-12 at 12 42 11 pm


peacecorps/peacecorps/static/peacecorps/css/src/components/_bubble.scss
```diff
@@ -78,35 +78,19 @@
 
 // TODO :not(old) is a hack to ignore styles in old browsers
 .bubble_form {
-  label {
-    cursor: pointer;
-
-    span {
-      @include font-size(1.9);
-    }
+  .small {
```
msecret: There shouldn't be modifiers like this, they should follow bem syntax like:

.bubble_form--small.

and not be nested at all.

That being said, the bubble class as numerous specificity problems right now that have to be refactored, so I'm fine leaving this as a TODO.

peacecorps/peacecorps/static/peacecorps/js/src/jsusers.js
```diff
@@ -0,0 +1,15 @@
...
+
+var $ = require('jquery');
+
+var showHideForJSUsers = function() {
+  $('.js-showForJSUsers').attr('aria-hidden', false);
+  $('.js-hideForJSUsers').attr('aria-hidden', true);
+};
+
+module.exports = {
+  showHideForJSUsers: showHideForJSUsers
```
msecret: I'm confused by the name of this. Is it showing for JS users or hiding for JS users?

peacecorps/peacecorps/static/peacecorps/js/src/jsusers.js
```diff
@@ -0,0 +1,15 @@
+/*
+ * Show parts of a page which would otherwise be hidden if JS is present
+ */
+'use strict';
+
+var $ = require('jquery');
+
+var showHideForJSUsers = function() {
```
msecret: This also hasn't been unit tested. It might help to pass in a $root to the function, which allows for easier testing.

PR #224, opened 2015-01-12
Removed "All Rights Reserved" from Footer

seanherron:
Proposing that we remove "All Rights Reserved" from the footer. This is for a number of reasons:
1. It's ambiguous. What rights are we referring to? This is normally (incorrectly) used in conjunction with copyright, so we can just assume that.
2. It's meaningless. Unlike trademarks, where the work must be aggressively protected in order to retain the trademark, copyright does not need to be expressly stated for a copyright to take effect.
3. The Peace Corps doesn't hold copyright to the content of the website, so it's factually incorrect to state that rights are reserved. As a work of the United States Government, content on peacecorps.gov is in the public domain. Particular rights may be embedded in statutes for things like the seal (22 U.S. Code § 2518 deals with the Peace Corps seal in particular), but that's a bit different than stating "All Rights Reserved" across the entire site.

### DIFF
--- a/peacecorps/peacecorps/templates/donations/base.jinja
+++ b/peacecorps/peacecorps/templates/donations/base.jinja
@@ -158,10 +158,7 @@
         <li class="sub_nav__link">
           <a class="t-body--sm"
              href="http://www.peacecorps.gov/about/policies/foia/">
-            Freedom of Information Guide</a>
-        </li>
-        <li class="sub_nav__link t-body--sm">
-          All Rights Reserved.
+            Freedom of Information</a>
         </li>
       </ul>
     </nav>


kgallagher:
These are good points, but we will have to have GC weigh on this. Yes, what we produce is public domain, but there are different issues related to photography shared with us by PCVs. We had concerns from GC when we tried to remove this during development of the mobile site. For now, we should keep it.

seanherron:
@kgallagher Good point regarding photography. Peace Corps GC should write specific language for objects which have differing copyright restrictions. Stating "All Rights Reserved" does not actually convey any meaningful information to the user.

kgallagher:
How do other agencies handle this? Would be helpful to have some examples.

seanherron:
Here's an example from CFPB that @cmc333333 pointed out: http://www.consumerfinance.gov/privacy-policy/

It states:

> No copyright may be claimed for any work on this website that was created by a federal employee in the course of his or her duties. Images and text appearing on this website are in the public domain and may be freely copied, unless otherwise noted. Credit is requested. If copyrighted material appears on the website, or is reached through a link on this website, the copyright holder must be consulted before the material may be reproduced.

As the conditions of copyright change depending on the circumstances of the asset in question, normally it is provided inline with the asset itself. For instance, something under CC BY 4.0 (http://creativecommons.org/licenses/by/4.0/) simply requires attribution, but other conditions may require additional restrictions.

PR #225, opened 2015-01-12
Small css refactoring

msecret:
- stop importing util in specific files
- remove donation-form
- move certain imports around, will change to be more structured after.


cmc333333:
Is there a corresponding source change for `js/compiled/peacecorps-donation.js`?

msecret:
Hmm no, I may have been bad and forgotten to compile previously.

PR #227, opened 2015-01-13
Resize: finish sorter

msecret:
Finshes resizing of the sorter.

Jinja cleanup
- Combined separate campaign macros into one.

CSS cleanup
- Gets rid of references to deprecated grid class and section__content
- Refactors column class to be used with any grid system.
- Adds more specific things to discover.

@ericronne you  may want to check how things look. It should also show up on the dev site after this is closed

screen shot 2015-01-12 at 4 52 30 pm
screen shot 2015-01-12 at 4 52 43 pm
screen shot 2015-01-12 at 4 52 53 pm
screen shot 2015-01-12 at 4 53 13 pm
screen shot 2015-01-12 at 4 53 25 pm


cmc333333:
Something's not right -- would you mind pulling from upstream?

PR #242, opened 2015-01-15
404 + Cleanup

cmc333333:
- Custom 404
- Removes blocks which are no longer needed; replaces with "title" context in all views
- Removes paths + templates which are no longer used
- Removes #amount-form from links on the landing/sorter pages

### DIFF
--- a/peacecorps/peacecorps/templates/404.jinja
+++ b/peacecorps/peacecorps/templates/404.jinja
@@ -1,0 +1,14 @@
+{% extends "donations/base.jinja" %}
+{% set title="Page Not Found" %}
+
+{% block content %}
+<section class="blurb blurb--sm">
+  <div class="t-blurb--md t--black">Error</div>
+  <div class="t--black">
+    The page you're looking for is unavailable. Sorry for the inconvenience.
+    Please try one of the links below instead.
+  </div>
+</section>
+{% endblock %}
+
+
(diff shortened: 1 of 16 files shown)


peacecorps/peacecorps/templates/404.jinja
```diff
@@ -0,0 +1,13 @@
+{% extends "donations/base.jinja" %}
+
+{% block content %}
+<section class="blurb blurb--sm">
+  <div class="t-blurb--md t--black">Error</div>
```
msecret: This should probably be an h1 rather then a div.

peacecorps/peacecorps/templates/404.jinja
```diff
@@ -0,0 +1,13 @@
+{% extends "donations/base.jinja" %}
+
+{% block content %}
+<section class="blurb blurb--sm">
```
msecret: If you put the t--black class here, you won't have to repeat it twice below ( I think).

peacecorps/peacecorps/templates/404.jinja
```diff
@@ -0,0 +1,13 @@
...
+  <div class="t-blurb--md t--black">Error</div>
+  <div class="t--black">
```
msecret: This could be a p or a h4.

cmc333333:
Change to h1/h4. Forgot a screenshot the first time. Note that this is the 404 page for donations alone, so we're using the donations header/footer.

screen shot 2015-01-16 at 9 38 32 am

PR #243, opened 2015-01-16
Resize header

msecret:
Resized the header and made styled the logo very specifically.

Other changes:
- Fixed some styling on the footer.


peacecorps/peacecorps/static/peacecorps/css/src/base/_type.scss
```diff
@@ -67,6 +67,7 @@ body {
 }
 .t--gray--2 {
   color: $gray-2;
+  text-decoration-color: $gray-2;
```
cmc333333, line 70: Looks like this isn't in Chrome or IE: http://caniuse.com/#feat=text-decoration 

Just to confirm, it's not needed in those browsers (and it makes things right in FF/Safari)?

peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -40,21 +40,21 @@
   </nav>
   <header class="banner banner--support banner--no_line section__content">
     <a href="{{ url('donate landing') }}">
-      <img src="{{ static('peacecorps/img/pc_logo.png') }}" class="u-inline"
+      <img src="{{ static('peacecorps/img/pc_logo.png') }}" class="u-inline
+          logo__img"
```
cmc333333: If giving this its own class, would it make sense to move those inline styles into the style sheet? It'd make resizing based on media queries easier, yes?

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -33,13 +33,14 @@
 </nav>
 
 <article class="js-discoverApp discover">
-  <ul class="js-filterable u-unmarked_list u-clearfix"
+  <ul class="js-filterable u-unmarked_list u-clearfix" id="issue"
```
cmc333333, line 36: Adding ids to these lists will cause the footer links ("By Issue", "By Country", "By Volunteer") to skip over the sorter's header. Just to confirm -- that's intended, yes?

peacecorps/peacecorps/templates/donations/donate_all.jinja
```diff
@@ -240,8 +243,8 @@
 {%- endmacro %}
 
 {% macro render_project(project, ptype) -%}
-  <article class="js-collapsibleItem u-clearfix drift_corner"
-        id="collapsible-{{ ptype }}-{{ project.id }}" aria-hidden="true">
+  <article class="js-collapsibleItem .js-hideForJSUsers u-clearfix drift_corner"
```
cmc333333: Stray period

cmc333333: Also, could you do the same for country and sector funds? Otherwise they, too, cannot be reached by non-JS users.

peacecorps/peacecorps/static/peacecorps/css/src/modules/_logo.scss
```diff
@@ -0,0 +1,33 @@
...
+ * Logo
+ */
+
+@import '../mixins';
+
+.logo {
+  margin-top: .45em;
+  .logo__text {
+    @include font-size(3.5);
+    font-family: 'Georgia';
```
cmc333333: Their current site uses this for font: `georgia,times,"times new roman",serif`. Should all these fallbacks be included here, as well?

peacecorps/peacecorps/static/peacecorps/css/src/base/_type.scss
msecret, line 70: This was only needed in firefox, so I think its fine to have here.

peacecorps/peacecorps/templates/donations/base.jinja
msecret: Yea good point, I didn't see the inline styling, so good catch!

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret, line 36: Yea I think its worth to keep this for a good experience for non-JS users as well as screen readers.

peacecorps/peacecorps/templates/donations/donate_all.jinja
msecret: They were already not being hidden by default, but were being hidden with JS code. I checked the page without JS, and see issue funds, country funds and projects.

peacecorps/peacecorps/static/peacecorps/css/src/modules/_logo.scss
msecret: Yea thats probably safest.

PR #244, opened 2015-01-16
Revised sitewide notification banner to match mockup

seanherron:
@msecret apologies if you have already had plans for this, had a bit of spare time so I decided to take a whack at doing some front end to help out. Delete this if it is redundant or the approach doesn't match how you would do it.

Before:
screen shot 2015-01-16 at 12 34 41 pm

Now:
screen shot 2015-01-16 at 12 35 30 pm


seanherron:
Merged in compiled CSS from `master` and updated

PR #245, opened 2015-01-16
Adjusted "Peace Corps" text in header

seanherron:
Brings in line with mockups to show "Peace Corps" in Georgia and slightly bigger.

Before:
screen shot 2015-01-16 at 1 04 27 pm

After:
screen shot 2015-01-16 at 1 04 41 pm


seanherron:
Closing this as a dupe

PR #258, opened 2015-01-27
Ajax for donation submission

cmc333333:
- Adds endpoints for submitting donor info. These will 400 on validation errors and 200 with paygov info
- Adds "full" project check to the form as well (it used to just live in views)
- When a user with JS clicks submit, their browser will try the API. If that fails (due to invalid data, a network error, etc.), the user's browser submits the form as normal. If it is successful, the paygov info is used to submit to pay.gov
- Tests for both the api and the JS


seanherron:
This corrects the JS jump issues we were seeing earlier on confirmation pages, yes?

cmc333333:
@seanherron I believe so. Before, all users went to the confirmation page and JS users had the form immediately submit. This caused a lot of confusion, particularly when the connection was slow.

This PR doesn't affect non-JS users (who still see the confirmation page). JS users, though, will submit the first form via AJAX, and, if successful, immediately POST to pay.gov, skipping the confirmation page altogether. If there is an error when they submit, the page will refresh as with non-JS users.

PR #259, opened 2015-01-27
508 change some gray links

msecret:
Icon text should be the dark gray so is 508 compliant.

Fixes https://github.com/18F/peace-corps-product-management/issues/332


cmc333333:
Yay variables :)

PR #260, opened 2015-01-28
ehnc link project/campaign titles.

msecret:
part of sorter adjustments
https://github.com/18F/peace-corps-product-management/issues/330


cmc333333:
Makes sense and matches the orange=link motif.

PR #263, opened 2015-01-28
Added login warning to admin page

seanherron:
Okay, so this solves https://github.com/18F/peace-corps-product-management/issues/209 but I'm not sure if there is a more elegant way to handle it. Basically, I'm overwriting the entire `admin/login.html` template. Extending the template didn't seem to provide much utility, as then I would have had to extend the base template as well, at least if I understand non-app admin template changes properly. Thoughts?


cmc333333:
This type of solution works. Those templates aren't very easy to extend -- @annalee found similar problems when changing the CSS.

PR #264, opened 2015-01-28
Faq Styles

seanherron:
This brings FAQ styling in line with mockups.

Current:
screen shot 2015-01-28 at 4 12 48 am

Mockup:
image

New:
screen shot 2015-01-28 at 4 13 32 am

Note I didn't include open styles in the same way as the mockup to keep things simple - that can be a future fix.


cmc333333:
Looks like inclusion of the compiled JS was unintentional?

PR #266, opened 2015-01-30
Changed special funds page to memorial funds page, moved global fund donate to sorter page

seanherron:
I think I caught every instance of special funds and changed it to memorial funds. Note that the bottom bar to go _back_ to the sorter was removed as there is a regression in the CSS that is also causing a bug on the current `master` branch.

screen shot 2015-01-30 at 10 24 38 am


cmc333333:
Looks good to me. Excited to see it

PR #267, opened 2015-01-30
Replaced language on memorial page

seanherron:
Realized the language I put in wasn't actually from Peace Corps, but instead from an external organization. Replacing this until we have official language to use.


msecret:
Even stuff like this should eventually be put into the CMS system.

PR #269, opened 2015-01-30
Make memorial images in line with mockups

seanherron:
This PR changes the styling of the memorial page to be more in line with the mockups. I'm using a few inline styles I want to optimize, but didn't want to substantially edit the SCSS as @msecret has a structure set up that I didn't want to interfere with.

screen shot 2015-01-30 at 6 56 21 pm


msecret:
Wouldn't this mean that the actual fund text won't be shown? I think Eric's mocks required a text section to come up when you click on a fund.

ericronne:
Don't forget the 3-pixel stroke around the circles
#AAA698

Also add space above the circles (which are larger than the design, but that's okay because i believe the client wanted to give these funds a little more prominence on the page)...
special funds - adj

cmc333333:
Something to keep in mind, here: there are now 9 memorial funds.

cmc333333:
Closing as this is quite stale.

PR #274, opened 2015-02-02
Ehnc styleguide

msecret:
Add a live style guide to static section

This is an auto generating css styleguide that can serve as a documented
reference and allow for a visual test environment.

It uses node-kss to find comments in the css files to generate the
guide.

There is both a guide and a template checked in. The template is written
while the actual guide is generated.

screen shot 2015-02-02 at 3 25 13 pm

screen shot 2015-02-02 at 3 25 25 pm


cmc333333:
Beautiful

PR #282, opened 2015-02-03
Ehnc styleguide modules

msecret:
Added modules to styleguide


cmc333333:
This adds two more `pc_logo.png`s. Is there a way to just reference the existing image?

msecret:
I'm mixed on this. For the styleguide, I feel like if you're developing for it and want to include an image to show something, you can just copy it into the styleguide dir and reference it. Another thing would be to have some sort of function, like the python static function, that somehow references the original. This would be a pain to build though.

I do see this being  a problem down the road if there are multiple images that have to be referenced in the styleguide, although based on how this site design is going now, I'm not sure that's going to be a problem now.

Another thing: I'm wondering if it will be useful for the styleguide to be able to exist on its own, ie not requiring any files ../ from itself.

PR #284, opened 2015-02-04
Build compile css

msecret:
Adds a build process for compiling the css to minified files and added conditionals in the template file to load minified files only in production.

Other changes:
- Refactored gruntfile to use variables for paths rather then strings.


peacecorps/peacecorps/settings/dev.py
```diff
@@ -68,6 +68,8 @@
 MEDIA_ROOT = './media/'
 MEDIA_URL = '/media/'
 
+JINJA2_CONSTANTS['DEBUG'] = True
```
cmc333333: Would you mind using something more descriptive, like "MINIFIED"? "DEBUG" is misleading.

msecret: What about something like ENV = n where n = settings file. That would mean this code could be used in cases that are not just minification.

msecret: But then it's confusing on which one would be base. I just went with minified for now.

PR #298, opened 2015-02-10
Read more link

cmc333333:
- Adds a "read more" link to projects/funds whose abstract has been shortened
- Makes the funding button link to the donation amount form

screen shot 2015-02-10 at 4 21 23 pm
screen shot 2015-02-10 at 4 21 37 pm


msecret:
I would add more spacing to the left but I'll merge this anyway.

PR #306, opened 2015-02-19
Content Updated

seanherron:
Updated with some new content from Peace Corps


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -128,6 +128,9 @@
             <li class="sub_nav__link">
               <a href="{{ url('donate memorial funds') }}">Memorial Funds</a>
             </li>
+            <li class="sub_nav__link">
+              <a href="/donate/fund/peace-corps">Peace Corps Fund</a>
```
cmc333333: `{{url ('donate campaign', slug='peace-corps')}}`

PR #314, opened 2015-02-23
Initial Global Site Configuration Proposal

seanherron:
This PR is a proposal. We probably shouldn't merge it in until we have a solid deployment strategy. Basically, this introduces a "Site Configuration" concept, which enables us to have one-off singleton settings that can apply sitewide. Here's what it looks like on the admin panel:

screen shot 2015-02-23 at 4 02 28 pm

screen shot 2015-02-23 at 4 03 29 pm

To accomplish this, you have to create a model in `site_configuration.models` that looks like this:

```
class SitewideNotification(SingletonModel):
    enable = models.BooleanField(
        help_text="Enables the Sitewide Announcement.",
        default=False)
    header = models.CharField(
        max_length=100,
        help_text="The Header Text for the current site-wide announcement.")
    body = models.TextField(
        max_length=255,
        help_text="The body text for the site-wide announcement, viewable on first page load.")
```

Then, add it to `admin.py`:

```
admin.site.register(models.SitewideNotification, SingletonModelAdmin)
```

Then, you can access it in templates via the following:

```
<h2>{{ get_config('SitewideNotification').header }}</h2>
```

Some outstanding questions:

1) How do we manage this well so there aren't millions of settings all over the place?
2) Is this the best (technical) approach to solving the problem?
3) How can I write better tests for this? The test as is right now is failing due to an SQL error I can't fully figure out:

```
django.db.utils.OperationalError: no such table: site_configuration_name
```

4) This probably will increase DB queries quite a bit if deployed aggressively. What's the best way to cache things?

### DIFF
--- a/peacecorps/site_configuration/models.py
+++ b/peacecorps/site_configuration/models.py
@@ -1,0 +1,31 @@
+from django.db import models
+
+from peacecorps.models import Media
+
+class SingletonModel(models.Model):
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        self.pk = 1
+        super(SingletonModel, self).save(*args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        pass
+
+    @classmethod
+    def get_obj(cls):
+        obj, created = cls.objects.get_or_create(pk=1)
+        return obj
+
+class Donate(SingletonModel):
+    intro_text = models.TextField(
+        max_length=255,
+        help_text="Introductory Text at the top of the donations landing page.")
+    about_donating = models.TextField(
+        max_length=255,
+        help_text="Text about how donating supports projects.")
+    featured_image = models.ForeignKey(
+        Media, null=True, blank=True,
+        help_text="A large landscape image for the top of the page")
--- a/peacecorps/site_configuration/models.py
+++ b/peacecorps/site_configuration/models.py
@@ -28,4 +28,20 @@ class Donate(SingletonModel):
         help_text="Text about how donating supports projects.")
     featured_image = models.ForeignKey(
         Media, null=True, blank=True,
-        help_text="A large landscape image for the top of the page")
+        help_text="A large landscape image for the top of the page",
+        related_name='donate_landing_featured_image')
+    sorter_featured_image = models.ForeignKey(
+        Media, null=True, blank=True,
+        help_text="A large landscape image for the top of the sorter page",
+        related_name='donate_sorter_featured_image')
+
+class SitewideNotification(SingletonModel):
+    enable = models.BooleanField(
+        help_text="Enables the Sitewide Announcement.",
+        default=False)
+    header = models.CharField(
+        max_length=100,
+        help_text="The Header Text for the current site-wide announcement.")
+    body = models.TextField(
+        max_length=255,
+        help_text="The body text for the site-wide announcement, viewable on first page load.")
--- a/peacecorps/site_configuration/templatetags/get_singleton.py
+++ b/peacecorps/site_configuration/templatetags/get_singleton.py
@@ -1,0 +1,29 @@
+from django import template
+from django.db import models
+from django.utils.translation import ugettext as _
+
+from django_jinja import library
+import jinja2
+
+register = template.Library()
+
+@register.assignment_tag(name='get_site_configuration')
+@library.global_function
+def get_singleton(model_path):
+    try:
+        app_label, model_name = model_path.rsplit('.', 1)
+    except ValueError:
+        raise template.TemplateSyntaxError(_(
+            "Templatetag requires the model dotted path: 'app_label.ModelName'. "
+            "Received '%s'." % model_path
+        ))
+    model_class = models.get_model(app_label, model_name)
+    if not model_class:
+        raise template.TemplateSyntaxError(_(
+            "Could not get the model name '%(model)s' from the application "
+            "named '%(app)s'" % {
+                'model': model_name,
+                'app': app_label,
+            }
+        ))
+    return model_class.get_singleton()
--- a/peacecorps/site_configuration/templatetags/get_singleton.py
+++ b/peacecorps/site_configuration/templatetags/get_singleton.py
@@ -7,17 +7,9 @@ import jinja2
 
 register = template.Library()
 
-@register.assignment_tag(name='get_site_configuration')
 @library.global_function
-def get_singleton(model_path):
-    try:
-        app_label, model_name = model_path.rsplit('.', 1)
-    except ValueError:
-        raise template.TemplateSyntaxError(_(
-            "Templatetag requires the model dotted path: 'app_label.ModelName'. "
-            "Received '%s'." % model_path
-        ))
-    model_class = models.get_model(app_label, model_name)
+def get_config(model_name):
+    model_class = models.get_model('site_configuration', model_name)
     if not model_class:
         raise template.TemplateSyntaxError(_(
             "Could not get the model name '%(model)s' from the application "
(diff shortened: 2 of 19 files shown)


cmc333333:
I really think this is a bad precedent to set. This is (was?) similar to Drupal's architecture, which had a big "variables" table, which was loaded on every page. The thinking was, this would be flexible enough to contain any extra junk the developers wanted to put in it. This was quickly and completely abused, and caused tremendous overhead and confusion. 

Effectively it's delineating content into two buckets, a CMS and a big bag of "configuration". As we progress, "configuration" will have more and more requirements -- for example, this is already missing the ability to write markup -- until there are effectively two CMSes: one for the "main" content and one for all the "auxiliary" content. We need a better long-term strategy. I don't know if that means a highly tuned sir trevor, an in-line editor, or what, but I don't think it looks like this.

gboone:
> 1) How do we manage this well so there aren't millions of settings all over the place?

I'm not sure you can. Once you set the precedent that this is the "extra stuff model" it's the extra stuff model. This seems a little easier to maintain (or at least notice once it's gotten out of hand) than similar solutions like the one @cmc333333 mentione because you'll have a growing list of `admin.site.register(models.SitewideNotification, SingletonModelAdmin)` lines in `admin.py` and at some point somebody will notice it's getting a bit heavy. Based purely on my limited experience with Django I would imagine you're going to have a lot of fun(?) with migrations, too, if you ever try to get rid of these settings. And if you don't get rid of them then you have bunch of useless stuff in the Django admin. 

> 2) Is this the best (technical) approach to solving the problem?

On the one hand, this _is_ content. On the other, it's not content that's going to be changed often (ever?) and presumably will go away someday. My thought is, keep the admin for content you know you want people editing on arbitrary timelines. Django gives you a flexibility for managing stuff like this through templates. This also has the advantage of being version controlled so if it _does_ change you have a better sense of who changed it and how. Plus, content creators and editors shouldn't have to deal with a bunch of settings they don't (and don't need to) understand, another thing most CMSs haven't really nailed.

gboone:
In defense of @seanherron and the Pull Request, though, you're not going to have a million settings for _your_ implementation here. To the extent you're setting the "extra stuff model" precedent you're doing it for your team and anybody who decides to take this up as a fork in the future. For _your_ use, you're not going to want to change the date format, timezone, permalinks, etc. the kind of stuff that typically ends up in the `variables` table or the Settings -> General screen. Somebody else might pick up this as an open source project and go nuts adding all that crap in but it doesn't seem likely that anyone on this development team would take it that far.

If there are are a relatively small set of extra things that don't fit into a better model, the "extra stuff model" might work. I'd think hard on whether these extra things can fit somewhere else, or whether there's a more descriptive model that could be used, before embarking, though.

seanherron:
So, I think we need to find a good middle ground here. Namely, I don't think requiring a developer to branch code, make a change, commit, make a pull request, have that PR accepted, and re-deploying the site is a feasible solution, especially around something like the landing image on the homepage, which probably may change with some regularity.

As @gboone said, we can be intelligent about how we _introduce_ settings so as to avoid the mess of having way too many, and organize this well from the beginning.

annalee:
I worked on a previous project where I handled ephemeral content like this with a single 'textblock' model. I think the fields were title, body, and notes. We'd use the body for whatever text was actually appearing on the page, and the title to label the block (notes was for comments between maintainers about what was in the block and why). Then we'd select the particular block we wanted with textblock.objects.get(title="foo") and display it that way.

This saves having to create and register a new model every time you need a semi-ephemeral text block, but it still needs to be implemented carefully to keep it from turning into a hacky mess.

cmc333333:
I wonder if the key-value approach works from a technical perspective and we just need a nice UI for the admins, perhaps broken down by the urls to said content.

annalee:
What I wanted to do and never got implemented on the project in question was a thing where logged-in admin users would get an option to edit the text inline when looking at the page on which it appears, in a manner similar to how linkedin lets you edit your profile.

seanherron:
So, I'm not sure I see the value of the key-value approach over managing this via controlled models. This is for a few reasons:
1. By having at least some barrier of entry here (making a new model) we can control the growth of the settings page. I'm worried that just having this be something you create in admin+templates will quickly grow even more crazy, as @annalee noted. 
2. How would we manage non-text entries on this? One of the reasons I like the singleton approach is that models can be URLFields, Media foreign keys, etc.
3. How would that realistically look in admin? Is having a single page full of key-value pairs going to be better than having some crafted admin pages?

cmc333333:
I'm referring to your approach as a key value ;)

seanherron:
@cmc333333 how so? I would think storing the key and value inside a model (like @annalee's textblock model) is the definition of a kv store?

cmc333333:
The approach you've put forth is to have a class instead, one which is dynamically looked up at render time. That plus a field name is your key. Your values must adhere to specific schemas, but the schemas have no meaning outside of the individual context. The key determines the semantics of the value, hence I see little difference from a key-value blob store. You could think of them as validated documents, if you prefer.

That gets to core of my message, though: it seems like the admin interface will be the key to making this work. I recommend it be broken down by page that the content will be seen on, preferably with an in-place editor like Annalee described. At that point it doesn't matter to the end user if you're validating Python models, JSON schema, or SQL inserts.

peacecorps/site_configuration/models.py
```diff
@@ -0,0 +1,47 @@
...
+        help_text="Introductory Text at the top of the donations landing page.")
+    about_donating = models.TextField(
+        max_length=255,
+        help_text="Text about how donating supports projects.")
+    featured_image = models.ForeignKey(
+        Media, null=True, blank=True,
+        help_text="A large landscape image for the top of the page",
+        related_name='donate_landing_featured_image')
+    sorter_featured_image = models.ForeignKey(
+        Media, null=True, blank=True,
```
cmc333333, line 34: Both images should probably be required, hey?

peacecorps/site_configuration/tests/tests.py
```diff
@@ -0,0 +1,18 @@
...
+
+class SigletonTest(TestCase):
+
+    def setUp(self):
+        self.template = Template(
+            "{{ get_config('Name').site_name }}"
+        )
+
+    def renders_site_config(self):
+        Name.objects.create(site_name='Test Config')
```
cmc333333, line 16: The test fails because you're trying to insert something (a `Name` model) which doesn't have an associated table. Just refer to an existing model, like `SitewideNotification`.

peacecorps/site_configuration/templatetags/get_singleton.py
```diff
@@ -0,0 +1,21 @@
...
+from django.utils.translation import ugettext as _
+
+from django_jinja import library
+import jinja2
+
+register = template.Library()
+
+@library.global_function
+def get_config(model_name):
+    model_class = models.get_model('site_configuration', model_name)
```
cmc333333, line 12: This will cause the "max number of queries" test to fail on the sorter, as each call to get_config will make another DB query. What about caching the retrieved models (django has a local memory cache), or doing this manually in the template with a single `set`?

khandelwal:
I think we need to step back and think about what the problem this is trying to solve actually is. Is it that the content changes often? Is it that the content needs to be changed without a developer in the loop? Or that we just need a place to put the content that's not in the template?

Also, an interesting way to think about this would be: what does all this look like if you want to release a spanish language version of the site? 

Content management is a really hard problem to solve.

cmc333333:
Content management is indeed a hard problem; I think that's why we've avoided a full blown, generic CMS (with the ability to translate, user-definable models, group blogs, etc. etc.). There's a lot of weight associated with that solution, both technically and conceptually.

I read the problem that this PR is trying to address as "there is content (particularly text strings and images) which certain users need to update relatively frequently, not surfaced in their interface".

cmc333333:
Closing this as it's quite out of date

PR #317, opened 2015-02-24
Reduce sorter weight by simplifying JS and markup

cmc333333:
Until this point, the "pages" in the sorter were represented more or less by the same structures in both JS and in the markup. This proved to be a major problem as the live sorter clocks in at around 1.4 M -- we know that more than half of this is redundant project information.

This PR reworks that page to be ran via filters. Controllers are given `data-controls-filter` attributes and filtered elements are given `data-in-filter` attributes. The latter can be in _multiple_ filters by comma-separating the filter names. In the markup, all projects list which country pages and issue pages they should be shown. This means that the projects only appear once in the markup, but can be displayed through multiple controllers.

The back "show all issues"/etc. button is handled by a very simple history stores as an array in the `state` field. This history also lets us highlight the selected sorts.

Also renamed some things.

Locally, this reduces the page size by about a half. With more projects, this savings should be improved.


peacecorps/peacecorps/static/peacecorps/css/compiled/donation.css
```diff
@@ -67,7 +67,7 @@ body {
   border-color: #A12122; }
 
 [aria-hidden=true] {
-  display: none; }
+  display: none !important; }
```
seanherron, line 70: I'd be curious about @msecret's thoughts on introducing `!important` in to the CSS. All past experience of mine has been that important tags open up a whole can of worms in terms of unexpected behaviors, but this may be scoped narrowly enough to not be a big deal.

cmc333333, line 70: You raise a good point. Previously, the sorter used `show()/hide()`s; when I replaced these with `aria-hidden`, elements stopped disappearing. This was because a class had a `display: block` property which was binding tighter than the aria-hidden attribute. I figured, aria-hidden _should_ bind tighter (in the general case), so I added the important flag. 

Other options include
- edit just that one (`display: block`) class -- I think that'd lead to similar confusing cases in the future
- use `show/hide` -- I read somewhere that this is actually pretty slow; it certainly has the same confusion factor

msecret, line 70: I'm fine with using !Important in this specific instance.

In CSS, !important is a powerful tool that can cause problems if you don't use it the correct way. Base on how this rule is being used, you should never have an aria-hidden=true element not be diplsay: none, so this rule is fine

PR #318, opened 2015-02-24
Add state to title of volunteer on third tab

cmc333333:
Looks like:
screen shot 2015-02-24 at 12 23 01 pm


kgallagher:
Looks great! Thank you!!

PR #325, opened 2015-02-27
Create sorter-specific classes to remove redundancy

cmc333333:
As the sorter repeats certain elements a few hundred times, this PR condenses the number of classes for most repeated elements on the page.

It reduces page size by about 30% (~100 projects) without changing functionality. Retested in chrome, safari, and ie11


seanherron:
I would merge this but would prefer @msecret to review first

cmc333333:
Not a super pressing concern; could easily be delayed until mid-late next week.

peacecorps/peacecorps/static/peacecorps/css/src/components/_rectangular_progress.scss
```diff
@@ -12,30 +12,6 @@
   width: 100%;
 }
 
-.rectangular_progress--mini {
```
msecret, line 15: Thanks for deleting unused code!

peacecorps/peacecorps/static/peacecorps/css/src/modules/_discover.scss
```diff
@@ -14,17 +14,230 @@
...
+  @extend .t-link--gray;
+}
+
+.discover_page--aside {
+  @extend .column;
+  @extend .column--third;
+  @extend .enclose--sm;
+  @extend .sm-hide;
+
+  div {
```
msecret: I'm worried about this rule.

In the case of the img below, its better because an image is more specific. But in the case of a div, it could be anything that this would select.

peacecorps/peacecorps/static/peacecorps/css/src/modules/_discover.scss
```diff
@@ -14,17 +14,230 @@
...
-      margin-bottom: -2em;
-      padding-top: 1.5em;
-    }
+.discover_button {
+  @extend .button;
+  @extend .button--sm;
+  @extend .button--primary;
+}
+
+.discover_progress {
```
msecret, line 195: One suggestion I would make is to limit nesting here and instead using namespacing.

so 

``` scss
.discover_progress

.discover_progress__bar
.discover_progress__fill
```

or with sass

``` scss
.discover_progress {
  &__fill {
```

If you feel like that will be too verbose, I'd be willing to let it go

cmc333333, line 195: How about an idiom like 

```
.discover_progress {
    dp__fill {
```

Underscores indicate it's a child of another selector, so the abbreviation may be enough?

msecret, line 195: yea thats fine by me

cmc333333:
Updated with your suggestions @msecret, thanks!

PR #328, opened 2015-03-02
Always use HTTPS for includes

seanherron:
There's no reason to use protocol-relative includes for assets. We can safely default to always using https. This also fixes some errors using protocol-relative URLs with opengraph tags.


peacecorps/peacecorps/templates/donations/base.jinja
```diff
@@ -25,7 +25,7 @@
         "peacecorps/css/compiled/donation.css") }}" />
     {% endif %}
     <link rel="stylesheet" type="text/css"
-        href="//cloud.typography.com/6638272/662206/css/fonts.css" />
+        href="https://cloud.typography.com/6638272/662206/css/fonts.css" />
```
khandelwal, line 28: Wouldn't you want this hosted locally? Instead of relying on a third-party server/service?

PR #342, opened 2015-03-06
Share tool on Donate

msecret:
Moved over all stylings, behavior and markup for the share tool.

Tested:
- Chrome
- Firefox
- Safari
- IE 11
- Iphone, both orientations
- IPad, both orientations

Scaled the beta panel down on mobile to make more room for the share too,
it also looks way better if it doens't drop to two lines.

Theres some overlap at a small amount of mobile sizes due to the page header
getting in the way. I think we should deal with this when we fix the header on
mobile.

screen shot 2015-03-05 at 4 53 09 pm
screen shot 2015-03-05 at 4 53 21 pm
screen shot 2015-03-05 at 4 54 01 pm
screen shot 2015-03-05 at 4 54 10 pm


cmc333333:
I think you should have access to the `request` in the templates if you want to get the url. You can create a global function (for all templates) to get the share data from settings (like `humanize_cents`, but without params). Note that on the success pages, the twitter/facebook/etc. all point to the project/campaign page (rather than the success page itself) -- not sure how that should be factored in. Let me know if I can help.

peacecorps/peacecorps/templatetags/share.py
```diff
@@ -0,0 +1,22 @@
...
+from django_jinja import library
+
+
+@library.global_function
+def share_tweet():
+    return getattr(settings, 'SHARE_TWEET', '')
+
+@library.global_function
+def share_url(request, override=None):
+    if override is not None:
```
cmc333333: Shouldn't this be `if override is None:`?

cmc333333:
Excellent

PR #346, opened 2015-03-11
Compiled JS in master was out of date

cmc333333:


cmc333333:
Going to merge as this is just a missing build step.

PR #353, opened 2015-03-26
Don't log donation if it's ACH. Add notice to success pages

cmc333333:
plus test

Looks like:
screen shot 2015-03-26 at 10 25 00 am
screen shot 2015-03-26 at 10 24 07 am


seanherron:
:+1:

PR #356, opened 2015-03-26
Update Docs + Add Quickstart Vagrant

seanherron:
This PR aims to simplify the onboarding process for a new developer. Namely, https://github.com/18F/peacecorps-site/issues/352 raises some great issues that we needed to improve on.

By using Vagrant, and including some stuff that was previous in a different repository, standing up a dev environment is now:

```
$ vagrant up
$ vagrant ssh
$ python manage.py runserver 0.0.0.0:8000
```

Issue #352, opened 2015-03-25
Getting setup is maybe harder than it should be?


phildini:
Hello! I have trying to make my way through some of the 'help wanted' issues, and have been unable to get a working devenv setup. It seems like the development flow is better optimized for linux, and I imagine that's where the app is being deployed, so would the maintainers of this project be adverse to me setting up a Vagrantfile and shell script for bootstrapping a devenv?

seanherron:
Thanks for reaching out, @phildini! We'd definitely be interested in a pull request for a working Vagrant development environment and appreciate your willingness to help. If you're looking for some inspiration, here's what I'm using in a different django project elsewhere in 18F:

https://github.com/18F/tock/blob/master/provision/dev/bootstrap.sh
https://github.com/18F/tock/blob/master/Vagrantfile

cmc333333:
And of course, if you're having any specific problems, post here and we'll try to work them out and update the README.

phildini:
I'm going to try and port the Vagrantfile soon, but for the record here's what I went through before giving up installing on a mac.
1. Installed python3 through homebrew.
2. Installed mkvirtualenv
3. ran `mkvirtualenv --python=/usr/local/bin/Python3 peacecorps-site`
4. mkvirtualenv failed with `OSError: [Errno 62] Too many levels of symbolic links`
5. Tried pyenv method, same mkvirtualenv problem.
6. Used `pyvenv` to create a virtualenv in the peacecorps-site directory - successfully created
7. Activated virtualenv, tried to install requirements.txt, got the following error:

```
    In file included from src/_pylibmcmodule.c:34:
    src/_pylibmcmodule.h:42:10: fatal error: 'libmemcached/memcached.h' file not found
    #include <libmemcached/memcached.h>
             ^
    1 error generated.
    error: command 'clang' failed with exit status 1
    ----------------------------------------
    Cleaning up...
```

cmc333333:
Good point. We use the memcached library (specifically, the production settings use this). In debian/ubuntu land, this would be an `apt-get install libmemcached-dev`. I assume homebrew is similar?

phildini:
The saga in getting setup continues. Is it helpful to continue documenting issues here or should I move on?

cmc333333:
Closing, per comments in #358

PR #357, opened 2015-03-26
A start on a vagrant dev environment.

phildini:
Right now everything works, but could be set up to be more
user friendly through the use of automatic aliases, etc.

Also, a "dev" settings set that can be used out of the box but allows
debug.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,64 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = "2"
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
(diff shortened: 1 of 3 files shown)


phildini:
This is also a starting point on #352

seanherron:
Thanks so much, @phildini! I actually expanded on your suggestions and rewrote the entire `README` as well as customized the Vagrantfile and provisioning script to include some fixtures and data specific to this repository (we merged that literally 3 minutes before you opened this PR, see https://github.com/18F/peacecorps-site/pull/356). I don't think there are any enhancements you did here that aren't present in that PR, can you take a look & try things out to see if this helps?

phildini:
I'll go look right now. Thanks for jumping on it!

seanherron:
One quick note - you may want to pull from https://github.com/18F/peacecorps-site/pull/358 instead, as I removed a nodejs dependency that was messing things up.

phildini:
Hah! I just ran into that.

PJJ
http://philipjohnjames.com

PR #358, opened 2015-03-26
remove nodejs

seanherron:
Removes node.js from the dev build as I wasn't able to get it working right and accidentally committed it.


phildini:
FWIW, I was able to start from scratch and get a working dev env with this pull request. I have no authority here, but if I did I would say :ship: 

Thank you!

phildini:
Also, this closes #352 when it's merged. :)

cmc333333:
We'll probably need to add something similar in as without it, script/style changes can't be made.

Thanks for prodding and testing @phildini !

Issue #98, opened 2014-11-17
Rename Sass support color


msecret:
Rename from $primary_support to just $support.

PR #363, opened 2015-04-21
Change "$primary_support" to "$support".

cliftonmcintosh:
This change is for https://github.com/18F/peacecorps-site/issues/98


cmc333333:
Thanks @cliftonmcintosh ! I'll add a follow up PR with the built css.

PR #364, opened 2015-04-21
Update css map file to latest

cmc333333:
Follow up to @cliftonmcintosh 's #363


cmc333333:
I'm going to merge this as it's not just a final pipeline step for #363

PR #365, opened 2015-04-28
Custom Open Graph Image for LGL

sshami:
This change involves implementing a custom open graph image for when users share the Let Girls Learn fund on Facebook. For now, I’ve just added a conditional statement in the og_image template block that looks for the LGL account code and overrides the og_image based on that. In the future, if this becomes a repetitive request to have custom open graph images for different funds, may need to think of a better solution.

Even though the open graph metadata is changing successfully, I wasn’t able to actually test the Facebook share feature on my local since FB isn’t able to scrape the locally hosted site. After merging, could we deploy to Staging first and test the share feature there before deploying to Production?

Thanks


cmc333333:
Nice

cmc333333:
@sshami this is up on the demo site now. Unless you say otherwise, it'll get to production this evening.

cmc333333:
@sshami Sorry, missed this yesterday. Following through tonight.

sshami:
@cmc333333 Ok, thanks. I won't be able to actually test the Facebook share till it's on Production since the OG url is set to the production url. OG metadata however is updating as expected on LGL fund in Staging.

PR #369, opened 2015-06-09
Paygov banner alert for June 13

sshami:
Hi @cmc333333 , we will need to put up another Pay.gov alert on Saturday June 13th and taken off Sunday morning. If it's more convenient for you to deploy it Friday night and take it off Monday morning, that's fine as well. This can be deployed the same way the previous pay.gov alert was deployed (as a tagged release of the paygov-banner-alert branch). Let me know what works for you. Thanks!

### DIFF
--- a/peacecorps/peacecorps/templates/donations/checkout_form.jinja
+++ b/peacecorps/peacecorps/templates/donations/checkout_form.jinja
@@ -1,5 +1,7 @@
 {% extends "donations/checkout_base.jinja" %}
 
+{% include "donations/includes/paygov-alert.jinja" %}
+
 {% block form_content %}
     <form action="" method="post" class="js-form section__content"
           data-ajax-url="{{ajax_url}}">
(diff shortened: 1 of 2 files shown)


cmc333333:
Can do -- I'll merge this into a separate branch tomorrow.

Something to consider - it seems like paygov might be requiring this more than we originally assumed. Probably worth writing a model so that this can be added in the UI?

khandelwal:
Having done things like this before, something consider: 

Build a model that has a start_timestamp and an end_timestamp for the message. Only display the message when we're in the timestamp window. That way the message is deactivated automatically, instead of requiring human intervention.

sshami:
Thanks @cmc333333 - that's something we will consider if these request continue. @khandelwal Thanks for the suggestion, seems like a good automated solution.

cmc333333:
@khandelwal what a good _idea_ [;)](https://github.com/cfpb/idea-box)

cmc333333:
@sshami this is on the staging site now. Unless you tell me otherwise, it'll go to prod in within that window.

cmc333333:
Went to prod Sat morning, removed Sun evening

PR #371, opened 2015-06-24
Paygov Alert Model

sshami:
Hi @cmc333333 

There is another pay.gov service interruption scheduled for this weekend (Sat June 27 6pm - midnight). Since this seems to be very recurring, I went ahead and created a model for pay.gov alerts as suggested so we could setup the alerts through the admin panel when needed. Can we possibly get this up before the weekend to do some testing on Staging? Let me know how you plan to deploy or if you have any questions regarding this change. Thanks!

Note/FYI: There was one small styling addition I made in the "type" sass file and ran the build which is why there's changes to some compiled static files.  Also, I was considering a singleton model for this, but since some consider that bad practice, it's setup as a simple model and the view checks the latest record in the table.


peacecorps/peacecorps/templates/donations/checkout_form.jinja
```diff
@@ -1,5 +1,9 @@
 {% extends "donations/checkout_base.jinja" %}
 
+{% if pay_gov_alert.is_active %}
```
cmc333333, line 3: Won't this fail when `pay_gov_alert` is `None` (as when there is no instance in the db)?

cmc333333:
Merging and pushing to staging, but there is a question about a potential bug

peacecorps/peacecorps/templates/donations/checkout_form.jinja
cmc333333, line 3: I must be mistaken -- no explosion on staging

sshami:
@cmc333333 Thanks! I had the same concern about the None object being called in the template but it seems like there were no complaints so I let it be. The changes look good in Staging, let me know when you can deploy to Production.

cmc333333:
@sshami I'll aim for tonight

cmc333333:
Pushed to prod

sshami:
@cmc333333 Looks good, thanks!

PR #372, opened 2015-07-16
Fixed Swaziland Map and Updated Fixture

sshami:
Hi @cmc333333 , 
When you get the chance, could you merge and deploy this. The Swaziland country code (SWZ) was incorrectly set to Switzerland's ISO 3166-1 code (CHE) in the fixtures - therefore showing Switzerland on it's country map. I also modified the Swaziland map svg a bit.

Since we will have to change the code for Swaziland in the admin panel for country data that's live now, I took a quick look to make sure 'sync accounting' was using country name based matching to determine the country, rather than country code based matching - so I don't think we should have an issue changing the country code for Swaziland in the admin panel. Let me know otherwise.

Thanks


cmc333333:
:+1:

PR #374, opened 2015-08-12
Remove donor name param from success page

sshami:
Hi @cmc333333 ,
When you get the chance, could you merge and deploy this to the dev site. The change involves removing the donor name from the redirect urls after the pay.gov transaction. This was a slight concern due to our GA tracking donor first names in the url. The django tests passed (after a bit modifications to the test since the redirect_urls method's donor_name param was remove), but we'd also like to test the full donation process on dev before this is deployed to production.

Thanks,
Saad

### DIFF
--- a/peacecorps/peacecorps/payxml.py
+++ b/peacecorps/peacecorps/payxml.py
@@ -143,18 +143,16 @@ def generate_custom_fields(data):
     return custom
 
 
-def redirect_urls(account, donor_name):
+def redirect_urls(account):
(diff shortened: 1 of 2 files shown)


cmc333333:
@sshami this is up on the demo site now

sshami:
@cmc333333 thanks! works fine. It can go up on production whenever you get the chance.

cmc333333:
And live!

PR #377, opened 2015-09-30
Google tag manager

sshami:
@defbyte I'm adding the pull request right now, but merging and deploying this should hold off until the Staging environment is setup since a QA process needs to happen beforehand to make sure new Google Tag Manager tracking is working. Thanks.


sshami:
@jacobgodin It may be beneficial to track failures - but currently we've just been asked to track successful donations in order to get an idea of which projects/campaigns are being donated to the most. In the future, we're also hoping to be able to track donation amount - but that's another conversation.

sshami:
@defbyte Let me know if you suggest doing a rebase for this pull request or if merging as is would be fine. Again, waiting on Staging environments to be setup next week, then we can merge and deploy to Staging.

PR #431, opened 2016-07-06
Pin Pillow to 3.0.0

defbyte:
This should resolve our build problem


sshami:
@defbyte Thanks! Was just about to ask about this.

defbyte:
ok @sshami this takes care of the build issue. I highly recommend we consider pinning additional requirements, because right now they can change major version number 🎲

